PR #1, opened 2022-10-06
PPF-29 Create persistence layer for Integration creation

LucWollants:
### Note
- The only purpose of this pull request is to create a simple domain and persistence layer for an "Integration"
- On top of this pull request, the Laravel Nova setup will be built
- Validation is still minimal but will be handled with Laravel Nova Resource validation, that is also why several of the domain classes use simple strings instead of value objects

### Added
- Added an abstract `UuidModel` to make Laravel Eloquent work with a UUID as primary key
- Added persistence layer for `Organization`
  - For now this is a standalone model until `User` is fully implemented
  - Which has one `Address`
- Added persistence layer for `Contact`
  - Which belongs to an `Integration`
- Added persistence layer for `Subscription`
  - For now this is a standalone model until `User` is fully implemented
- Added persistence layer for `Integration`
  - For now this is a standalone model until `User` is fully implemented
  - Which has many `Contacts` and one `Subscription`
- Added basic `Makefile`

### Changed
- Updated GitHub Actions with MySQL service and dedicated `.ci.env`

---


app/Domain/Integrations/IntegrationType.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace App\Domain\Integrations;
+
+enum IntegrationType: string
+{
+    case EntryApi = 'entry-api';
+    case SearchApi = 'search-api';
+    case UitpasApi = 'uitpas-api';
```
bertramakers, line 11: UiTPAS won't be included in the first phase if I remember correctly, so maybe best to leave this out for now?

app/Domain/Organizations/Models/OrganizationModel.php
```diff
@@ -0,0 +1,30 @@
...
+        'name',
+        'vat',
+    ];
+
+    /**
+     * @return HasOne<AddressModel>
+     */
+    public function address(): HasOne
+    {
+        return $this->hasOne(AddressModel::class, 'organization_id');
```
bertramakers, line 28: I'm not against putting `Address` in a separate model/table, but I'm not sure either why we'd do that if an organization has exactly one address? Wouldn't it be easier to put those 4 columns inside the `organizations` table? For example for:

- Quickly looking up the address of an organization in the table (e.g. for debugging)
- Manual queries on the organizations/addresses tables
- Exports of the organizations/addresses tables data
- Ensuring that an organization always has an address (right now it could not have a row in the addresses table)

The only advantage I see to putting it in a separate table right now is that we could make it a one to many relation later (so one organization could have multiple addresses), but that seems quite unlikely?

app/Domain/Integrations/IntegrationType.php
LucWollants: Agreed! Will do that.

app/Domain/Organizations/Models/OrganizationModel.php
LucWollants: Agreed! I will change it. I focussed too much on building a POC on how Laravel Nova works with relationships and less on what we need ☺️

app/Domain/Subscriptions/Subscription.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace App\Domain\Subscriptions;
+
+use Ramsey\Uuid\UuidInterface;
+
+final class Subscription
+{
+    public function __construct(
```
LucWollants, line 11: Add integration type relation, see https://github.com/cultuurnet/publiq-platform/pull/2#discussion_r1011642286

app/Domain/Subscriptions/Models/SubscriptionModel.php
```diff
@@ -0,0 +1,25 @@
...
+namespace App\Domain\Subscriptions\Models;
+
+use App\Models\UuidModel;
+use Illuminate\Database\Eloquent\SoftDeletes;
+
+final class SubscriptionModel extends UuidModel
+{
+    use SoftDeletes;
+
+    protected $table = 'subscriptions';
```
bertramakers, line 14: While reviewing #4 I realized that because we store the subscriptions in a table, we won't have any subscriptions locally. And we'll also have to keep the subscriptions on 3 environments in sync: acc, test and prod. (In theory it doesn't matter what the subscriptions are on acc/test, but for predictability it might be best to keep them the same as on prod.)

Maybe we should keep them in a config file instead so we can easily keep them in sync on all 4 environments? (local, acc, test, prod)

LucWollants: Or we could create a [seeder](https://laravel.com/docs/9.x/seeding) for that, which could become part of the deployment phase like the migrations.
I will create a seeder for now and we can finetune later when things are more concrete.

app/Domain/Subscriptions/Subscription.php
LucWollants: Added an `integrationType` to a `Subscription`

app/Domain/Subscriptions/Models/SubscriptionModel.php
bertramakers: Can we avoid the subscriptions being editable in Nova with a seeder? And won't a seeder be a bit complex for someone like @corneelwille to edit the subscription info when needed?

LucWollants: > Can we avoid the subscriptions being editable in Nova with a seeder?
We can make the subscriptions read-only in Nova

> And won't a seeder be a bit complex for someone like @corneelwille to edit the subscription info when needed?
Yes, but the seeder can be filled with values from a config.

I think this is something we can implement in a follow-up PR.

bertramakers: Okay sounds good. I only don't see why we need a database table, Eloquent model, and database seeder for subscriptions if they will be read-only and read from a config file in the end. 😄

bertramakers (approved):

PR #2, opened 2022-10-12
PPF-29 Setup laravel nova

LucWollants:
### Added
- Added Laravel Nova
- Created Nova Resource for User
- Created Nova Resource for Subscription


JonasVHG (approved):

app/Domain/Subscriptions/Subscription.php
```diff
@@ -13,9 +13,9 @@ public function __construct(
         public readonly string $name,
         public readonly string $description,
         public readonly Currency $currency,
-        public readonly int $price,
+        public readonly float $price,
         public readonly BillingInterval $billingInterval,
-        public readonly ?int $fee,
+        public readonly ?float $fee,
```
bertramakers, line 18: As discussed on Slack, Subscription / SubscriptionModel should also have an integration type column so we that when its linked to a new integration, we can filter out the subscriptions that are not meant for the integration's type.

public/vendor/nova/app.css
```diff
@@ -0,0 +1,3 @@
...rem}.md\:py-5{padding-bottom:1.25rem;padding-top:1.25rem}.md\:py-0{padding-bottom:0;padding-top:0}.md\:py-3{padding-bottom:.75rem;padding-top:.75rem}.md\:py-8{padding-bottom:2rem;padding-top:2rem}.md\:px-12{padding-left:3rem;padding-right:3rem}.md\:px-2{padding-left:.5rem;padding-right:.5rem}.md\:px-3{padding-left:.75rem;padding-right:.75rem}.md\:px-0{padding-left:0;padding-right:0}.md\:py-6{padding-bottom:1.5rem;padding-top:1.5rem}.md\:pt-2{padding-top:.5rem}[dir=ltr] .md\:pr-3{padding-right:.75rem}[dir=rtl] .md\:pr-3{padding-left:.75rem}.md\:text-2xl{font-size:1.5rem;line-height:2rem}.md\:text-xl{font-size:1.25rem;line-height:1.75rem}.md\:text-sm{font-size:.875rem;line-height:1.25rem}.md\:text-\[4rem\]{font-size:4rem}.md\:font-normal{font-weight:400}}@media (min-width:1024px){.lg\:absolute{position:absolute}.lg\:top-\[56px\]{top:56px}.lg\:bottom-auto{bottom:auto}[dir=ltr] .lg\:ml-60{margin-left:15rem}[dir=rtl] .lg\:ml-60{margin-right:15rem}.lg\:block{display:block}.lg\:hidden{display:none}.lg\:w-60{width:15rem}.lg\:max-w-lg{max-width:32rem}.lg\:break-words{overflow-wrap:break-word}.lg\:px-6{padding-left:1.5rem;padding-right:1.5rem}.lg\:px-12{padding-left:3rem;padding-right:3rem}}
```
bertramakers, line 1: Can we ignore all these "vendor" files? (`public/vendor`, `lang/vendor`, ...)

app/Nova/Resources/Resource.php
```diff
@@ -0,0 +1,27 @@
...
+
+declare(strict_types=1);
+
+namespace App\Nova\Resources;
+
+use Illuminate\Database\Eloquent\Builder;
+use Laravel\Nova\Http\Requests\NovaRequest;
+use Laravel\Nova\Resource as NovaResource;
+
+abstract class Resource extends NovaResource
```
bertramakers, line 11: What added value does this abstract class have exactly? It seems like it overrides some methods but either implements them exactly the same way (`indexQuery`) or just calls the parent method (`detailQuery`, `relatableQuery`)

app/Domain/Subscriptions/Subscription.php
LucWollants: Add IntegrationType on Subscription Nova Resource

public/vendor/nova/app.css
LucWollants: Those files are added by running the `nova::install` command, see `https://nova.laravel.com/docs/4.0/installation.html

I took the approach of only requiring `composer install` to be run, but it can be changed of course.

@bertramakers what do you prefer?

app/Nova/Resources/Resource.php
LucWollants: Indeed, that can be removed and added again if needed. It's part of the default scaffolding when running `nova::install`

So I removed that class for now.

PR #4, opened 2022-10-25
PPF-29 Add React form with Inertia response to create an Integration

LucWollants:
### Added
- Added form to create an Integration


resources/js/Pages/Index.jsx
```diff
@@ -1,11 +1,14 @@
...
 const Index = () => {
   return (
     <div>
-      <h1>Index Page</h1>
+      <h3>Index Page</h3>
     </div>
   )
 };
 
+Index.layout = page => <Layout children={page} />;
```
simon-debruijn, line 12:
```suggestion
Index.layout = page => <Layout>{page}</Layout>;
```

The same `children={` is used 4 times in this PR

Passing children is almost always done by passing it into the jsx brackets. It's just a convention

bertramakers (commented):
I'm going to wait a bit to review this completely until the other PRs are finalized and merged, as some feedback on those can have some impact on the work in this PR

LucWollants:
> I'm going to wait a bit to review this completely until the other PRs are finalized and merged, as some feedback on those can have some impact on the work in this PR

@bertramakers everything is on main and also merged in the pull request. So ready for review again.

bertramakers (approved):
I took a quick glance at it but I don't think I can give a lot of feedback on this. If I'm not mistaken this is just a first shot at the form for new integrations which will be reworked by @brampauwelyn and/or @simon-debruijn anyway, and they will probably give feedback on the setup if it gives problems at that point. So I think we can just merge this as-is right now as a proof-of-concept and we'll have a ticket to style/rework the form later and see if the setup requires any changes then.

PR #7, opened 2022-11-07
Login with UiTiD through auth0

LucWollants:
### Added
- Added `auth0/login` package
- Added route `login` and `logout`
- Added route `auth/callback`

### Changed
- Changed `auth` to also accept `auth0`
- Changed `UserModel` to also implement `StatefulUser`

### Notes
- Currently, only the admin pages from Laravel Nova are protected. When visiting pages at `/admin` as an unauthenticated user a redirect will happen to the `/login` page.
- To trigger a manual login directly visit `/login`
- To trigger a manual logout directly visit `/logout`

---
Ticket: https://jira.uitdatabank.be/browse/PPF-65


routes/web.php
```diff
@@ -20,6 +23,11 @@
 
 Route::get('/', static fn () => Inertia::render('Index'));
 
+Route::get('/login', Login::class)->name('login');
+Route::get('/admin/login', static fn () => redirect('/login'));
+Route::get('/logout', Logout::class)->name('logout');
```
bertramakers, line 28: When I log out from the admin dashboard, I noticed I get redirected back to the UiTID login form. In theory this makes sense because after the logout the user is sent back to the admin page, and a login is required there so they get sent back to UiTID to login. However in reality this will probably be a bit confusing. Is there a way to just send them to a page that confirms that they're logged out and a link to login back again?

And/or, we could create a custom landing page for the admin zone with a login button like the landing page on https://www.uitdatabank.be (when you're not logged in) and show that on `/admin` when not logged in? So then when someone logs out and they see that page again it's probably clear enough that they're logged out. (Optionally we could show a message there confirming that they're logged like in the first suggestion)

routes/web.php
```diff
@@ -20,6 +23,11 @@
...
+Route::get('/auth/callback', Callback::class)->name('auth.callback');
```
bertramakers, line 29: Are these `->name()` calls required for the routes to work correctly with Laravel Nova and/or the Laravel Auth0 SDK, or are they optional? If they're optional I'd leave them out at this point because it could be unclear later if anything actually uses them.

config/auth0.php
```diff
@@ -0,0 +1,59 @@
...
+    'redirectUri' => env('AUTH0_REDIRECT_URI', env('APP_URL') . '/callback'),
+
+    // Client Secret, found in the Auth0 Application settings.
+    'clientSecret' => env('AUTH0_CLIENT_SECRET'),
+
+    // One or more API identifiers, found in your Auth0 API settings. The SDK uses the first value for building links. If provided, at least one of these values must match the 'aud' claim to validate an ID Token successfully.
+    'audience' => \Auth0\Laravel\Configuration::stringToArrayOrNull(env('AUTH0_AUDIENCE')),
+
+    // One or more scopes to request for Tokens. See https://auth0.com/docs/scopes
+    'scope' => \Auth0\Laravel\Configuration::stringToArray(env('AUTH0_SCOPE')),
```
bertramakers, line 33: We'll need to request specific scopes when fetching client access tokens to integrate with the management API (`read:clients`, `create:clients`, etc), and different ones when letting a user log in (`profile` and/or `refresh_token` and/or `email`). Can we make that distinction with this SDK?

config/auth.php
```diff
@@ -110,4 +115,12 @@
...
     'password_timeout' => 10800,
 
+
+    'login_parameters' => [
+        'locale' => 'nl',
+        'referred' => 'publiq-platform',
+        'prompt' => 'login',
+        'skip_verify_legacy' => 'true',
+        'product_display_name' => 'publiq platform',
+    ],
```
bertramakers, line 125: Is there a way to store these in a config file that will be stored inside the `cultuurnet/appconfig` repo instead? Since that is the easiest place for non-developers to configure them

routes/web.php
LucWollants: They are indeed optional, so I removed them.

config/auth.php
LucWollants: When storing an object it is indeed required to have multiple env variables to match all the properties. This was done here: https://github.com/cultuurnet/publiq-platform/blob/246556222043248ac1301381048f1cde7a28c29e/config/insightly.php

That makes it possible for non-developers to configure the object inside the env file. But that is not a solution for adding and removing login parameters.

The config file is pure PHP so we can always create a helper method to parse other files and take the values from other files than just `.env` files.

bertramakers: Yep we can add another PHP file for the login parameters and include that. Just a bit sad how we then end up with a PHP (app)config file + an ENV (app)config file, instead of just one PHP (app)config file or one ENV (app)config file. It's like a weird compromise to use an ENV file but also use a PHP file for the things that the ENV file cannot solve. 😅 But it's a nit and fine for me.

routes/web.php
LucWollants: This required some sorcery 😅

To make this work the default `admin/logout` page needs to be intercepted and execute the `Auth0` logout code.
But this still results in an issue because the default behavior is a redirect to the login page after logout.

The workaround for this issue is to return a response from the logout endpoint, which is a `POST` endpoint. When this response has a `redirect` property the JS code of Laravel Nova will do a redirect to the route specified in this property instead of the login route.

This is the Laravel Nova JS code that handles the logout response
```
    async attempt() {
      if (confirm(this.__('Are you sure you want to log out?'))) {
        this.logout(Nova.config('customLogoutPath'))
          .then(redirect => {
            if (redirect !== null) {
              location.href = redirect
              return
            }

            Nova.redirectToLogin()
          })
          .catch(e => {
            Inertia.reload()
          })
      }
    },
```

config/auth0.php
```diff
@@ -0,0 +1,59 @@
...
+declare(strict_types=1);
+
+/**
+ * Please review available configuration options here:
+ * https://github.com/auth0/auth0-PHP#configuration-options.
+ */
+return [
+    // Should be assigned either 'api', 'management', or 'webapp' to indicate your application's use case for the SDK.
+    // Determines what configuration options will be required.
+    'strategy' => env('AUTH0_STRATEGY', \Auth0\SDK\Configuration\SdkConfiguration::STRATEGY_REGULAR),
```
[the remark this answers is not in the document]
LucWollants: I tried changing `strategy` to `STRATEGY_MANAGEMENT_API` and this gives indeed issues when doing the login. So we will need to instantiate a separate instance of the Auth0 SDK to execute the management API calls.

config/auth0.php
LucWollants: We will need to instantiate a separate instance of the Auth0 SDK to execute the management API calls.

config/auth.php
LucWollants: In theory, we could work with config files only.

The only requirement for the `.env` file is the MySQL database values used by Docker to create a container with a pre-created database. We could workaround this by creating a custom image with predefined values.

For now, I'm gonna keep it as in the scope of this ticket, but I created https://jira.uitdatabank.be/browse/PPF-70 to do some extra investigation

routes/web.php
bertramakers: Looks good!

config/auth.php
bertramakers: Aha okay now I see why we'd need an `.env` file. In that case let's keep using that and just try to find a way to make the list of login parameters (both the keys & values) configurable by non-devs in the appconfig. For example with another separate PHP config file indeed.

bertramakers (approved):

PR #8, opened 2022-11-14
PPF-29 Add Insightly integration to create contact and opportunity

LucWollants:
### Added
- Added Insightly integration to create contact and opportunity

---


tests/Insightly/InsightlyClientTest.php
```diff
@@ -0,0 +1,53 @@
...
+        $contact = new Contact(
+            Uuid::uuid4(),
+            Uuid::uuid4(),
+            ContactType::Technical,
+            'Jane',
+            'Doe',
+            'jane.doe@anonymous.com'
+        );
+
+        $contactId = $this->insightlyClient->contacts()->create($contact);
```
steftrenson, line 51: Does this actually use a real call to insightly or is it mocked?

steftrenson (approved):

tests/Insightly/InsightlyClientTest.php
LucWollants: It is a real call. That is also why the delete was added. But for now, it's not part of the GitHub actions, maybe we can only run this on the main branch for example.

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -80,6 +81,8 @@ public function store(StoreIntegration $storeIntegration): RedirectResponse
 
         $this->integrationRepository->save($integration);
 
+        IntegrationCreated::dispatch($integrationId);
```
bertramakers, line 84: Maybe it's a bit better if the `IntegrationRepository` would dispatch these messages? Since the `save()` method on it could in theory be called from multiple places and not just this controller

app/Insightly/Json.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Insightly;
+
+final class Json
```
bertramakers, line 7: Does Laravel have a class for this by any chance? Or if not, can we move this to just `app` so we can re-use it in other places later and not just the Insightly integration?

bertramakers (commented):
Maybe it could be useful to have a way to disable the Insightly integration if there is no API key configured, so the app works when you do the local setup but are maybe waiting on someone to give you the API key? Or even so we don't spam Insightly from our local environments all the time but only when we actually need to work on that integration (so it's easier to find our opportunities / contacts etc).

app/Domain/Integrations/Controllers/IntegrationController.php
LucWollants: Yes! That is indeed a very good approach. Most Laravel examples use the Eloquent model directly so they suffer from this copy/paste issue. We have a central place to do the saving so we can do the dispatch there.

Note: the Laravel Nova code still needs the dispatch because it bypasses the repo.

app/Insightly/Json.php
LucWollants: Laravel doesn't have a class for this. I moved it to the root of the `App` namespace. If we add more JSON-related stuff it could get its own folder.

LucWollants:
> > Maybe it could be useful to have a way to disable the Insightly integration if there is no API key configured, so the app works when you do the local setup but are maybe waiting on someone to give you the API key? Or even so we don't spam Insightly from our local environments all the time but only when we actually need to work on that integration (so it's easier to find our opportunities / contacts etc).
> 
> Good suggestion. I could prevent the calls in `sendRequest` with an if-statement like this
> 
> ```
>         if (empty(config('insightly.api_key'))) {
>             return new Response(204);
>         }
> ```

This causes issues, so for now I put in an early return. Can be finetuned later and put in a more central place.

app/Domain/Integrations/Controllers/IntegrationController.php
bertramakers: As mentioned on Slack we could use the Eloquent events so it also works in Nova https://laravel.com/docs/9.x/eloquent#events

LucWollants: I refactored this to use the Eloquent events and it works smoothly. This is indeed the best approach to handle our domain events.

LucWollants:
I noticed a requirement is missing: "Adding the integration contacts to Insightly". But I think it is better to implement this in a separate pull request.

@steftrenson maybe something for you as an extra intro on the platform?

PR #9, opened 2022-11-15
Fix package name

bertramakers:
### Fixed

- Added a name to `package.json`. Otherwise the name in `package.lock` always gets overwritten to `html` when running `npm install`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
     "requires": true,
     "packages": {
         "": {
+            "name": "publiq-platform",
             "dependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -4,6 +4,7 @@
     "requires": true,
     "packages": {
         "": {
+            "name": "publiq-platform",
```
bertramakers, line 7: (Note that this one got added automatically after running `npm install`)

LucWollants (approved):

PR #10, opened 2022-11-16
Add email Simon de Bruijn to nova users config

simon-debruijn:
### Added

- Simon de Bruijn to nova users config

### DIFF
--- a/config/nova.php
+++ b/config/nova.php
@@ -91,6 +91,7 @@ return [
 
     'users' => [
         'dev@publiq.be',
+        'simon.debruijn@publiq.be',
     ],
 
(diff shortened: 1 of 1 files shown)


bertramakers:
On localhost you can login with any email address in Laravel Nova so you don’t need to add your email address to the list for that

bertramakers (approved):
@simon-debruijn clarified that this is to test Nova out on the staging server. I've approved it but we should look into a way to make this appconfig @LucWollants since we probably want to maintain a list per environment + non-devs will need to be able to maintain this (like the UDB3 god users)

PR #14, opened 2022-11-28
PPF-29 Upload contacts to Insightly

LucWollants:
### Added
- Added methods to upload contacts to Insightly
- Linked contact with a related opportunity in Insightly
- Added mapping table for internal UUID and external Insightly id

---


Makefile
```diff
@@ -29,6 +29,8 @@ stan:
 test:
 	vendor/bin/sail composer test
 
+ci: lint stan test
```
bertramakers, line 32: I think this must also be added to the `.PHONY:` line at the top?
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

database/migrations/2022_11_28_181600_create_insightly_mappings_table.php
```diff
@@ -0,0 +1,25 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('insightly_mappings', static function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->string('insightly_id');
+            $table->string('resource_type');
```
bertramakers, line 15: We probably need an index on this `resource_type` column?

PR #15, opened 2022-11-30
PPF-77 Bugfix migrated v1 accounts

JonasVHG:
### Changed

- Length of column `user_id` in `action_events`

### Fixed

- Users with a migrated v1 account(with `UserId` like `auth0|<<UUID>>` can save items without getting an error.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-77

### DIFF
--- a/database/migrations/2022_11_30_145724_update_action_events_table.php
+++ b/database/migrations/2022_11_30_145724_update_action_events_table.php
@@ -10,7 +10,7 @@ return new class () extends Migration {
     public function up(): void
     {
         Schema::table('action_events', function (Blueprint $table) {
-            $table->string('user_id', 42)->change();
+            $table->string('user_id', 255)->change();
         });
     }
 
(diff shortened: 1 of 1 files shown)


database/migrations/2022_11_30_145724_update_action_events_table.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('action_events', function (Blueprint $table) {
+            $table->string('user_id', 42)->change();
```
bertramakers, line 13: It's safer to make it `255` while you're at it, because we do not have control over the length of the Auth0 ids. For example when someone logs in via Google or Facebook their id is `google-oauth2|<random string>` or `facebook|<random string>`. We cannot expect the length to be the same for all those social logins forever, and more types of logins can be added in the future (for example Apple and Mijn Burgerprofiel were added recently)

bertramakers (approved):

PR #24, opened 2022-12-08
PPF-74 Setup horizon

LucWollants:
### Added
- Setup Horizon for async queueing and monitoring of Jobs

### Note
- The `QUEUE_CONNECTION` is set by default to `sync`. To experiment with this feature change `QUEUE_CONNECTION` to `redis`
- Extra information can be found here
  - [Laravel Horizon](https://laravel.com/docs/9.x/horizon)
  - [Laravel Queues](https://laravel.com/docs/9.x/queues)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-74


README.md
```diff
@@ -90,6 +90,27 @@ The following environment variables need to be set in the `.env` file
...
+
+By default Laravel processes jobs synchronously, because `QUEUE_CONNECTION` is set to `sync`.
+
+To enable Horizon, set the `QUEUE_CONNECTION` environment variable to `redis` in the `.env` file.
+
+Once asynchronous processing is enabled, you also need to start the queues with `make horizon`.
+
+### Monitoring
+
+Horizon is a dashboard for monitoring the queue. It can be accessed at [http://localhost/horizon](http://localhost/horizon).
```
bertramakers, line 105:
```suggestion
Horizon has a dashboard for monitoring the queue. It can be accessed at [http://localhost/horizon](http://localhost/horizon).
```

Unless it is "is" of course

simon-debruijn (approved):

app/Insightly/Listeners/CreateContact.php
```diff
@@ -11,11 +11,14 @@
 use App\Insightly\InsightlyMapping;
 use App\Insightly\Repositories\InsightlyMappingRepository;
 use App\Insightly\Resources\ResourceType;
+use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Support\Facades\Log;
 
 final class CreateContact implements ShouldQueue
 {
+    use Queueable;
```
bertramakers, line 20: Out of curiosity, what is the difference between `ShouldQueue` and `Queueable`?

LucWollants: In Laravel, every plain PHP class can be a "Job" and can be dispatched. But the `Queueable` interface is an indication for the Job Dispatcher to put this Job on the queue instead of dispatching it immediately.

The trait `Queueable` provides methods for the Job Dispatcher to interact with the queue. It is not required to make the queuing work. But it is common practice to add it from the start to make sure that methods like `onQueue` or `delay` will work when used during future development.

LucWollants: It is not very clear from the documentation.

Here is the info about the `ShouldQueue` interface https://laravel.com/docs/9.x/queues#generating-job-classes

The `Queueable` is not documented by itself, it is used in examples https://laravel.com/docs/9.x/queues#chain-connection-queue

LucWollants: Maybe another way to put it:

- Step 1: Make a job queueable by adding the interface `ShouldQueue` (internal Laravel code will now dispatch the job on a queue)
- Step 2: Need queue interaction then add the `Queueable` trait (but I think this name is misleading and should be `QueueInteractable`)

bertramakers: Thanks but I think what's confusing me is what "queue interaction" means exactly. What does `Queueable` do exactly that's different from the default behavior for queueable jobs (`ShouldQueue`)?

PR #26, opened 2022-12-09
PPF-29 Improve user info

LucWollants:
### Added
- Added first name and last name to custom `UserModel` with information retrieved from the `userinfo` endpoint

---

### DIFF
--- a/app/Domain/Auth/CurrentUser.php
+++ b/app/Domain/Auth/CurrentUser.php
@@ -28,6 +28,16 @@ final class CurrentUser
         return $this->user()->name;
     }
 
+    public function firstName(): string
+    {
+        return $this->user()->first_name;
+    }
+
+    public function lastName(): string
+    {
+        return $this->user()->last_name;
+    }
+
     private function user(): UserModel
     {
         /** @var UserModel $user */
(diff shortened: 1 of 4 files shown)


brampauwelyn (approved):

app/Domain/Auth/CurrentUser.php
```diff
@@ -28,6 +28,16 @@ public function name(): string
         return $this->user()->name;
     }
 
+    public function firstName(): string
+    {
+        return $this->user()->first_name;
```
bertramakers, line 33: Did you try this with a regular user (not a social login)? I think these fields are only filled in for social users in our case because UiTiD itself (the custom code on top of Auth0) does not store that info in these fields afaik. So we should probably not rely on them

LucWollants: I created an regular user and made the following changes to make it work for both type of users:
- Use `https://publiq.be/first_name` as the key to get the first name
- Add fallback to empty string when `last_name` is missing

bertramakers (approved):
For anyone following along: Apparently none of the "name" fields are 100% guaranteed to be present so we fall back on empty strings.

PR #28, opened 2022-12-09
PPF-83 Create Auth0 clients every time a new integration is created

bertramakers:
### Added

- Added event listener that creates an Auth0 client in every configured tenant for new integrations

### Changed

- Merged existing `AppServiceProvider` and `EventServiceProvider` into a new `InsightlyServiceProvider`
- Refactored `IntegrationRepository` to implement an interface

---
Ticket: https://jira.uitdatabank.be/browse/PPF-83


app/Auth0/Auth0ClientsForIntegration.php
```diff
@@ -0,0 +1,45 @@
...
+declare(strict_types=1);
+
+namespace App\Auth0;
+
+use InvalidArgumentException;
+use Ramsey\Uuid\UuidInterface;
+
+final class Auth0ClientsForIntegration
+{
+    private array $auth0ClientPerTenant = [];
```
simon-debruijn, line 12: I presume you can't type this array more specifically in php? 

Something like:

```php
private [ string => Auth0Client ] $auth0ClientPerTenant = [];
```

simon-debruijn: Or that is only possible with comment syntax like below?:

```php
 /**
     * @var Auth0TenantSDK[]
     */
```

app/Auth0/Auth0TenantSDK.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    public function createClientForIntegration(Integration $integration): Auth0Client
+    {
+        $name = sprintf('%s (id: %s)', $integration->name, $integration->id->toString());
+
+        $apis = match ($integration->type) {
+            IntegrationType::SearchApi, IntegrationType::Widgets => 'sapi',
+            IntegrationType::EntryApi => 'entry sapi',
+        };
```
simon-debruijn, line 39: Cool php has match expressions 🙂

simon-debruijn (approved):

app/Auth0/Auth0ClientsForIntegration.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function __construct(
+        public readonly UuidInterface $integrationId,
```
LucWollants, line 15: Would it be better to remove this redundant parameter `$integrationId` and make sure that all passed `Auth0Client` have the same integration id?

app/Auth0/Auth0ClientsForIntegration.php
```diff
@@ -0,0 +1,45 @@
...
+                        $auth0Client->integrationId->toString(),
+                        $this->integrationId->toString()
+                    )
+                );
+            }
+            $this->auth0ClientPerTenant[$auth0Client->tenant->value] = $auth0Client;
+        }
+    }
+
+    public function getClientForTenant(Auth0Tenant $tenant): ?Auth0Client
```
LucWollants, line 33: This method doesn't seem to be used.

bertramakers: Ah true, but we will need it later (for example in the frontend to show only the credentials for the test tenant for non-activated integrations) and it seems simple enough to keep?

app/Auth0/Auth0ClientsForIntegration.php
bertramakers: I initially did it like that but that doesn't work when an integration has 0 clients, for example when all tenant config is disabled (in tests or when working locally without tenant config for some reason) or when handling the `CreateClients` async. Because then this object will have no `$integrationId` at all

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -22,9 +22,9 @@
 final class IntegrationController extends Controller
 {
     public function __construct(
-        private readonly SubscriptionRepository $subscriptionRepository,
-        private readonly IntegrationRepository $integrationRepository,
-        private readonly CurrentUser $currentUser
+        private readonly SubscriptionRepository        $subscriptionRepository,
+        private readonly EloquentIntegrationRepository $integrationRepository,
```
LucWollants, line 26: Should this inject the interface?

I also see some auto-indentation from PHPStorm

app/Auth0/Auth0ClientsForIntegration.php
bertramakers: In PHPDoc it would have to be:

```
@var Auth0Client[]
```

(https://docs.phpdoc.org/latest/guide/references/phpdoc/types.html#arrays)

However then the type of the array keys is missing (since this is an associative array, meaning that the keys are strings instead of integers)

In Psalm it can be represented as:
```
@var array<string, Auth0Client>
```
https://psalm.dev/docs/annotating_code/type_syntax/array_types/

But I'm not 100% sure that PHPStan and other tools always work with that. Honestly I'm not sure it's worth the effort for such simple cases because it's not a native language construct, so the only thing we're adding is some kind of safety _depending on the tools we use_. Also because when PHP adds native support for this, it will probably not be by using `@` annotations.

app/Providers/InsightlyServiceProvider.php
```diff
@@ -4,12 +4,17 @@
...
 use App\Insightly\InsightlyClient;
+use App\Insightly\Listeners\CreateContact;
+use App\Insightly\Listeners\CreateOpportunity;
 use App\Insightly\Pipelines;
 use GuzzleHttp\Client;
+use Illuminate\Support\Facades\Event;
 use Illuminate\Support\ServiceProvider;
 
-final class AppServiceProvider extends ServiceProvider
+final class InsightlyServiceProvider extends ServiceProvider
```
LucWollants, line 17: To be consistent this can also be moved to the Insightly namespace

app/Domain/Integrations/Controllers/IntegrationController.php
bertramakers: I decided not to refactor classes that were out of scope to use the interface instead of the concrete class, I think that's something better suited for a separate PR (also to add interfaces to the other repositories)

app/Insightly/Listeners/CreateOpportunity.php
```diff
@@ -19,18 +19,14 @@ final class CreateOpportunity implements ShouldQueue
     use Queueable;
 
     public function __construct(
-        private readonly InsightlyClient $insightlyClient,
-        private readonly IntegrationRepository $integrationRepository,
-        private readonly InsightlyMappingRepository $insightlyMappingRepository
+        private readonly InsightlyClient               $insightlyClient,
+        private readonly EloquentIntegrationRepository $integrationRepository,
```
LucWollants, line 23: Maybe also the interface is required here? This function also has the same auto-indentation from PHPStorm

tests/TestCase.php
```diff
@@ -9,4 +9,22 @@
...
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->disableConfigExternalApis();
+    }
+
+    protected function tearDown(): void
+    {
+        $this->disableConfigExternalApis();
```
LucWollants, line 21: Is this needed? I would expect it only to be needed inside the `setUp` method

app/Auth0/Repositories/EloquentAuth0ClientRepository.php
```diff
@@ -0,0 +1,50 @@
...
+
+use App\Auth0\Auth0Client;
+use App\Auth0\Auth0ClientsForIntegration;
+use App\Auth0\Models\Auth0ClientModel;
+use Illuminate\Support\Facades\DB;
+use Ramsey\Uuid\UuidInterface;
+
+final class EloquentAuth0ClientRepository implements Auth0ClientRepository
+{
+    public function save(Auth0Client ...$auth0Clients): void
```
LucWollants, line 15: Would it make sense to change the type here to `Auth0ClientsForIntegration`?

app/Insightly/Listeners/CreateOpportunity.php
bertramakers: See https://github.com/cultuurnet/publiq-platform/pull/28#discussion_r1045616888

LucWollants (approved):
Looks good!

tests/TestCase.php
bertramakers: It felt safer but it's probably not needed indeed

app/Auth0/Auth0ClientsForIntegration.php
bertramakers: But this is not generics I think? The value type is always the same (`Auth0Client`) regardless of the instance of `Auth0ClientsForIntegration`

bertramakers: And again it's only for support in specific tools like PHPStan, which is not complaining in this case

app/Auth0/Auth0ClientsForIntegration.php
LucWollants: Does that also mean that the passed-in array of `$auth0Clients` is empty?

app/Auth0/Repositories/EloquentAuth0ClientRepository.php
bertramakers: I thought about it when refactoring it from saving one client to multiple. I decided to keep it generic, in the sense that the repository can save one or more `Auth0Client` regardless of how they were loaded/created. But I can change it if you prefer `Auth0ClientsForIntegration`. I'd rename the method then though

LucWollants: Keep it, it's fine.

app/Auth0/Auth0ClientsForIntegration.php
bertramakers: Yes, for example if you call `Auth0ClientRepository::getByIntegrationId()` and get 0 rows because the async command is not handled yet or there are no configured tenants, then it will return an instance of `Auth0ClientsForIntegration` with the given `$integrationId` and an array of 0 values for the clients. Since there are 0 rows/clients at that point, `Auth0ClientsForIntegration` cannot get the `$integrationId` from the client objects themselves so it needs to be injected in the constructor.

app/Auth0/Auth0ClientsForIntegration.php
LucWollants: That is because it is disabled here https://github.com/cultuurnet/publiq-platform/blob/main/phpstan.neon#L16

app/Auth0/Auth0ClientsForIntegration.php
bertramakers: I'm not sure I understand. If there are 0 rows the array of `$auth0Clients` will be empty, so the foreach won't do anything and the validation won't happen.

bertramakers: Maybe there is some confusion because `$integrationId` is not just an argument but also a property? It's just set automatically now using the `public readonly`?

LucWollants: So then I think that `$integrationId` is not needed as a parameter?

bertramakers: It's not just a parameter but a property (our messages probably crossed). I see now that it's unused out of the class though so I could remove it. But then I would rename the class to just `Auth0Clients` since the integration context is lost then

app/Providers/InsightlyServiceProvider.php
bertramakers: Done

tests/TestCase.php
bertramakers: Removed in [516f319](https://github.com/cultuurnet/publiq-platform/pull/28/commits/516f3197a562b455b637be8e08b665cdca2c2017)

app/Auth0/Auth0ClientsForIntegration.php
LucWollants: Keep it like it is, it's fine.

bertramakers: Update 2: I'm going to remove this class completely and just work with `array` 😄

LucWollants: Maybe a Laravel `Collection` is a good alternative?

app/Auth0/Auth0TenantSDK.php
bertramakers: @simon-debruijn yes but it's specifically for [matching possible values of enums](https://www.php.net/manual/en/control-structures.match.php). You are maybe comparing it to rust match expressions? In that case https://www.php.net/manual/en/control-structures.switch.php is more comparable I think

app/Auth0/Auth0ClientsForIntegration.php
bertramakers: At first glance that is more of a "util" class that provides chainable methods like `map()` etc so you don't need to use `array_map()`, and you can easily create it from an `array` when needed. So my instinct would be to just work with `array` to keep it a bit less coupled and use `collect()` or `new Collection()` where you want the helper methods. Or am I missing something like type checking? https://laravel.com/docs/9.x/collections

bertramakers: I've just refactored it to `array` now in [993c7bd](https://github.com/cultuurnet/publiq-platform/pull/28/commits/993c7bd1c049109125c07d421fa5a94b5a50e40b) and [0576d60](https://github.com/cultuurnet/publiq-platform/pull/28/commits/0576d60492fa4a44face023a16817a0f9c59d899) but if `Collection` is in fact the way to go it's easy to change.

LucWollants: I tried to return a `Collection` from the repositories. Make it also easy to work with in the HTTP Controller layer or even the Inertia views.

PR #37, opened 2022-12-22
PPF-78 Setup sentry

LucWollants:
### Added
- Setup Sentry SDK

### Changed
- Made sure `UserModel` can return a string for `id` attribute instead of an integer

---
Ticket: https://jira.uitdatabank.be/browse/PPF-78


steftrenson (approved):

bertramakers:
I just noticed that this breaks `make horizon` (and probably some other CLI commands as well):

```
bertramakers@Berts-MacBook-Pro publiq-platform % make horizon
vendor/bin/sail artisan horizon

   RuntimeException 

  Session store not set on request.

  at vendor/laravel/framework/src/Illuminate/Http/Request.php:559
    555▕      */
    556▕     public function session()
    557▕     {
    558▕         if (! $this->hasSession()) {
  ➜ 559▕             throw new RuntimeException('Session store not set on request.');
    560▕         }
    561▕ 
    562▕         return $this->session;
    563▕     }

      +9 vendor frames 
  10  app/Exceptions/Handler.php:50
      Illuminate\Support\Facades\Facade::__callStatic()

      +2 vendor frames 
  13  app/Exceptions/Handler.php:52
      Sentry\configureScope()
```

It looks like `configureScope()` tries to access the session, which requires an HTTP context. Is there a way to circumvent that in the CLI context?

PR #54, opened 2023-01-03
PPF-102 Create UiTiD v1 consumer(s) for new integration

bertramakers:
### Added

- Added event listener for `IntegrationCreated` that creates a consumer in every configured UiTiD v1 environment (acc/test/prod).

### Changed

- `Auth0ClusterSDK::createClientsForIntegration()` returns an `array` instead of `Collection`, because the only place where it was used I had to do an explicit `->toArray()` to get rid of a warning in PHPStorm. And since that means the only place that calls it expects an `array`, it makes more sense to just return that imho. (Noticed this while working on the UiTiD v1 equivalent.)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-102

### DIFF
--- a/app/UiTiDv1/Listeners/CreateConsumers.php
+++ b/app/UiTiDv1/Listeners/CreateConsumers.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\UiTiDv1\Listeners;
+
+use App\Domain\Integrations\Events\IntegrationCreated;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\UiTiDv1\Repositories\UiTiDv1ConsumerRepository;
+use App\UiTiDv1\UiTiDv1ClusterSDK;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+final class CreateConsumers implements ShouldQueue
+{
+    use Queueable;
+
+    public function __construct(
+        private readonly UiTiDv1ClusterSDK $clusterSDK,
+        private readonly IntegrationRepository $integrationRepository,
+        private readonly UiTiDv1ConsumerRepository $consumerRepository
+    ) {
+    }
+
+    public function handle(IntegrationCreated $integrationCreated): void
+    {
+        $integration = $this->integrationRepository->getById($integrationCreated->id);
+        $consumers = $this->clusterSDK->createConsumersForIntegration($integration);
+        $this->consumerRepository->save(...$consumers);
+    }
+}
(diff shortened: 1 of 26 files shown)


app/UiTiDv1/Listeners/CreateConsumers.php
```diff
@@ -0,0 +1,31 @@
...
+namespace App\UiTiDv1\Listeners;
+
+use App\Domain\Integrations\Events\IntegrationCreated;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\UiTiDv1\Repositories\UiTiDv1ConsumerRepository;
+use App\UiTiDv1\UiTiDv1ClusterSDK;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+final class CreateConsumers implements ShouldQueue
```
LucWollants, line 14: Maybe we should add some logging like is done for the Isnightly integration and also log failed jobs?
If so this can also be done for the Auth0 integration.

app/UiTiDv1/UiTiDv1EnvironmentSDK.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $formData = [
+            'name' => $name,
+            'description' => $integration->description,
+            'group' => $this->permissionGroupsPerIntegrationType[$integration->type->value] ?? [],
+        ];
+
+        $response = $this->sendPostRequest('serviceconsumer', $formData);
+
+        // Easiest way to convert XML to an array is by encoding an SimpleXMLELement as JSON and then decoding it again.
```
LucWollants, line 37:
```suggestion
        // Easiest way to convert XML to an array is by encoding an SimpleXMLElement as JSON and then decoding it again.
```

LucWollants (approved):
Nice work! Some complex domain code nicely abstracted behind simple interfaces.

app/UiTiDv1/Listeners/CreateConsumers.php
bertramakers: I'll take a look at how it's done for Insightly now. But if we also need to add it for Auth0, I think it's better to do it in a follow-up PR? (I exactly followed the approach of the Auth0 integration in this one)

Also, does Horizon not log the failed jobs?

LucWollants: Adding this in another pull request is perfect.

Horizon does indeed log it, but I thought it could be useful if we would at extra functionality on top of logging.
If for example, dev-ops set monitors on logging files everything is in a central place.

bertramakers: Alright, I've created a follow-up ticket for it: https://jira.uitdatabank.be/browse/PPF-106

bertramakers:
> Nice work! Some complex domain code nicely abstracted behind simple interfaces.

Mostly some complex OAuth1 HTTP requests haha

PR #57, opened 2023-01-04
PPF-87 Keep a history of model modifications

LucWollants:
### Added

- Database table to store histories
- History Model
- `Listener` to catch changes to Domain objects
- `MockUser` when `CurrentUser` is needed in tests

---
Ticket: https://jira.uitdatabank.be/browse/PPF-87


simon-debruijn (approved):

app/Domain/Histories/HistoryServiceProvider.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Support\ServiceProvider;
+
+final class HistoryServiceProvider extends ServiceProvider
+{
+    public function register(): void
+    {
+        $this->app->bind(HistoryRepository::class, EloquentHistoryRepository::class);
+
+        Event::listen(
+            'App\Domain\*',
```
bertramakers, line 20: If I'm not mistaken this only listens to the events that we explicitly define ourselves. But if someone would now update an integration (e.g. via Nova), it would not be recorded?

LucWollants: We will need a custom event for that to also make sure it gets updated inside Insightly.
More actions are now possible inside Nova than we currently handle.

bertramakers: I'm not sure every change to every entity type will result in an action in Insightly/Auth0/UiTiD v1. In those cases we might not create an event for it because there's no need for it, and then they won't be recorded in the history table.

bertramakers: Maybe this is a safer alternative, as it works for all Eloquent events? https://spatie.be/docs/laravel-activitylog/v4/introduction (specifically https://spatie.be/docs/laravel-activitylog/v4/advanced-usage/logging-model-events)

JonasVHG: I've tested this and only defined events are logged.
Having looked into https://spatie.be/docs/laravel-activitylog/v4/introduction I think it could be used as an alternative, but this also let's you clean the logs very easily, which might lead to accidental mistakes.

bertramakers: They have foreseen that:
```
If you want to automatically cleanup your production system you should append the --force option as the command will otherwise ask you to confirm the action. This is to prevent accidental data loss.
```

LucWollants: Maybe that package will make it harder to link a change to a certain user and to a certain model?
Because the link to a certain model is important to list this inside Laravel Nova.

app/Domain/Histories/EventToModelMapping.php
```diff
@@ -0,0 +1,25 @@
...
+use App\Domain\Integrations\Events\IntegrationCreated;
+use App\Domain\Integrations\Models\IntegrationModel;
+use App\Domain\Organizations\Events\OrganizationCreated;
+use App\Domain\Organizations\Events\OrganizationDeleted;
+use App\Domain\Organizations\Events\OrganizationUpdated;
+use App\Domain\Organizations\Models\OrganizationModel;
+
+final class EventToModelMapping
+{
+    public const MAPPING = [
```
LucWollants, line 19: Indentation is a bit weird here.

app/Domain/Histories/HistoryServiceProvider.php
bertramakers: I haven't tried it out, but the description on the homepage seems to cover that?

> The spatie/laravel-activitylog package provides easy to use functions to **log the activities of the users** of your app. It can also **automatically log model events**. All activity will be stored in the activity_log table.

(Emphasis mine)

It's possible that they don't store the user that triggered the model event, but given that their tagline is "Log the activities of your users" I assumed it would store that info as well for Eloquent models. I'd check the `activity_log` table to see if it has a user column, but I'll need to restart my computer first because Docker does not want to start 🤡 🐳

JonasVHG: I am making a quick setup to see if this works better. There is a `causer_id` which I assume would be the user.

bertramakers: Yes in the docs they have an example like `activity()->causedBy($user)` in the docs for logging your own custom activities. I'd guess that they also do that on Eloquent model events.

JonasVHG: `causer_id` is the user who performs it. 
If we go with this route We'll also have to use a custom migration like in https://github.com/cultuurnet/publiq-platform/blob/13c4815dd339d53b5abe32e4771ba71bac38be66/database/migrations/2022_11_30_145724_update_action_events_table.php#L13

bertramakers: You can use the provided migration to create the `activity_log` table as documented here: https://spatie.be/docs/laravel-activitylog/v4/installation-and-setup (I first confused it with the `action_events` table but it's a different one)

LucWollants:
Will be replaced with https://github.com/cultuurnet/publiq-platform/pull/72

PR #59, opened 2023-01-05
PPF-88 Update a contact in insightly

steftrenson:
### Added
- `ContactUpdated` event when a contact gets updated.
- `UpdateContact` listener that will sync an updated contact to Insightly

Todo:
- [x] Fix crash when contact wasn't synced to Insightly because it's a contributor

---
Ticket: https://jira.uitdatabank.be/browse/PPF-88


app/Domain/Contacts/Models/ContactModel.php
```diff
@@ -32,7 +42,10 @@ final class ContactModel extends UuidModel
     protected static function booted(): void
     {
         self::created(
-            static fn ($contactModel) => ContactCreated::dispatch(Uuid::fromString($contactModel->id))
+            static fn (ContactModel $contactModel) => ContactCreated::dispatch($contactModel->getId())
+        );
+        self::updated(
+            static fn (ContactModel $contactModel) => ContactUpdated::dispatch($contactModel->getId())
```
bertramakers, line 48: Nit: You could also do `$contactModel->toDomain()->id` and then you don't need the extra methods

app/Insightly/InsightlyServiceProvider.php
```diff
@@ -27,7 +28,9 @@ public function register(): void
     {
         $this->app->bind(InsightlyMappingRepository::class, EloquentInsightlyMappingRepository::class);
 
-        $this->app->singleton(InsightlyClient::class, function () {
+        $this->app->bind(InsightlyClient::class, HttpInsightlyClient::class);
+
+        $this->app->singleton(HttpInsightlyClient::class, function () {
```
bertramakers, line 33: What's the point of doing `bind()` and then `singleton()` with the implementation class name, if you could just do the `singleton()` definition with the interface name (like you did initially)?

bertramakers (approved):

app/Insightly/InsightlyServiceProvider.php
bertramakers: I personally prefer to just define the `singleton()` with the interface name, it's one less abstract layer that you need to resolve in your head when seeing how everything is wired. Only if the implementation can be auto-wired I'd use `bind()` because then you can just point out the implementation without implementing it yourself.

simon-debruijn (approved):

JonasVHG (approved):

app/Insightly/InsightlyServiceProvider.php
steftrenson: Well, this was a merge conflict as both Luc and I made a change. I prefer the direct binding as well instead of adding a step, but maybe @LucWollants had his reasons?

app/Domain/Contacts/Models/ContactModel.php
steftrenson: Good point, they shouldn't be used anyway, only from the domain model so it makes sense.

app/Insightly/InsightlyServiceProvider.php
LucWollants: I think the issue was that the `HttpInsightlyClient` was indeed created as a singleton. But `InsightlyClient` gets injected everywhere. So when running the application I got an error because `InsightlyClient` could not be created.

I quickly fixed that by binding the interface `InsightlyClient` to  `HttpInsightlyClient`. But now I notice I could also have changed the singleton key from `HttpInsightlyClient` to `InsightlyClient`.

So fine for me to just use the singleton with `InsightlyClient`

app/Insightly/SyncIsAllowed.php
```diff
@@ -0,0 +1,19 @@
...
+use App\Domain\Contacts\ContactType;
+
+final class SyncIsAllowed
+{
+    public static function forContact(Contact $contact): bool
+    {
+        return match ($contact->type) {
+            ContactType::Functional, ContactType::Technical => true,
+            ContactType::Contributor => false,
+        };
```
bertramakers, line 17: Very small nit: Since the return value is a boolean, you could also do it without a `match` like:
```
return $contact->type === ContactType::Functional || $contact->type === ContactType::Technical;
```

Or even:
```
return in_array($contact->type, [ContactType::Functional, ContactType::Technical]);

bertramakers (approved):

app/Insightly/SyncIsAllowed.php
steftrenson: I prefer the match statement as I find it more readable and works incredibly well with enums. An IDE can check whether you have matched all possible enum cases, which might also be added to a static analysis tool. This will detect potential issues when you add a new enum case, for instance.

PR #71, opened 2023-01-09
PPF-89 Delete contact

steftrenson:
### Added
- `ContactDeleted` event and associated `DeleteContact` listener to delete deleted contacts in Inisghtly.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-89

### DIFF
--- a/tests/Domain/Contacts/Repositories/EloquentContactRepositoryTest.php
+++ b/tests/Domain/Contacts/Repositories/EloquentContactRepositoryTest.php
@@ -6,7 +6,9 @@ namespace Tests\Domain\Contacts\Repositories;
 
 use App\Domain\Contacts\Contact;
 use App\Domain\Contacts\ContactType;
+use App\Domain\Contacts\Models\ContactModel;
 use App\Domain\Contacts\Repositories\EloquentContactRepository;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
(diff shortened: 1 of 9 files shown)


JonasVHG (approved):

bertramakers (approved):
Code changes look good, but maybe @LucWollants should chime in whether the behaviour is the same as in projectaanvraag and all the necessary checks are done. (In the ticket it's mentioned that the delete should only happen in some places, maybe? It's not clear to me.)

tests/Domain/Contacts/Repositories/EloquentContactRepositoryTest.php
```diff
@@ -106,4 +108,42 @@ public function test_it_can_get_all_contacts_from_an_integration(): void
...
+            'Doe'
+        );
+
+        $this->contactRepository->save($contact);
+
+        /** @var ContactModel $contactModel */
+        $contactModel = ContactModel::query()->findOrFail($contact->id);
+        $contactModel->delete();
+
+        $deletedContact = $this->contactRepository->getDeletedById($contact->id);
```
LucWollants, line 129: Nit: check that a normal `getById` throws?

LucWollants (approved):
Looks good to me and follows the approach of the other Insightly listeners.

PR #72, opened 2023-01-10
PPF-87 Log events via ActivityLog

JonasVHG:
### Added
- `ActivityLog` to store events

### Changed
- 

### Fixed

---
Ticket: https://jira.uitdatabank.be/browse/PPF-87


app/Domain/Integrations/Models/IntegrationModel.php
```diff
@@ -37,9 +37,12 @@ protected static function booted(): void
...
         );
-        self::deleted(
-            static fn ($integrationModel) => $integrationModel->update(['status' => IntegrationStatus::Deleted])
-        );
+    }
+
+    public function delete(): ?bool
+    {
+        $this->update(['status' => IntegrationStatus::Deleted]);
+        return parent::delete();
```
bertramakers, line 45: I'm curious, was this necessary to make something work in the ActivityLog library or was the change needed for some other reason? It's not possible to do it via the event anymore?

database/migrations/2023_01_09_134359_add_event_column_to_activity_log_table.php
```diff
@@ -0,0 +1,24 @@
...
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+final class AddEventColumnToActivityLogTable extends Migration
+{
+    public function up()
+    {
+        Schema::connection(config('activitylog.database_connection'))->table(config('activitylog.table_name'), function (Blueprint $table) {
+            $table->string('event')->nullable()->after('subject_type');
```
bertramakers, line 13: Is this not provided in the migration stub that creates the `activity_log` table? If not, is this column supported by out of the box by the activity log library? Because I don't seem to see any custom code that stores data in this extra column?

app/Domain/Integrations/Models/IntegrationModel.php
LucWollants: We moved the status code update to the `delete` method to ensure it no longer depends on the `deleted` model event being fired. When doing a `delete` the status should always be updated, it is not a side effect that should be handled when the `deleted` events triggers.

database/migrations/2023_01_09_134359_add_event_column_to_activity_log_table.php
LucWollants: I went through the migrations with Jonas and I proposed to create one simple migration that also uses the table name.

LucWollants:
As discussed @JonasVHG it can be useful to publish the config file.

database/migrations/2023_01_09_134359_add_event_column_to_activity_log_table.php
bertramakers: Ah I found it, it's a separate migration in the package itself https://github.com/spatie/laravel-activitylog/blob/main/database/migrations/add_event_column_to_activity_log_table.php.stub

That was not clear to me.

app/Models/UuidModel.php
```diff
@@ -37,6 +39,14 @@ public function getKeyType(): string
...
     public function getActivitylogOptions(): LogOptions
     {
-        return LogOptions::defaults();
+        return (new LogOptions())->logAll();
+    }
+
+    /**
+     * @return HasMany<Activity>
+     */
+    public function activityLog(): HasMany
```
LucWollants, line 48: Smart to place it here instead of the separate models!

app/Nova/Resources/ActivityLog.php
```diff
@@ -0,0 +1,60 @@
...
+final class ActivityLog extends Resource
+{
+    public static string $model = Activity::class;
+
+    public static $title = 'description';
+
+    /**
+     * @var array<string>
+     */
+    public static $search = [
```
LucWollants, line 30: Some weird indentation

app/Nova/Resources/ActivityLog.php
```diff
@@ -0,0 +1,60 @@
...
+                'subject_id',
+            ];
+
+    /**
+     * @return array<Field>
+     */
+    public function fields(NovaRequest $request): array
+    {
+        return [
+                ID::make()
```
LucWollants, line 41: Some weird indentation

LucWollants (approved):

app/Policies/ActivityPolicy.php
```diff
@@ -0,0 +1,54 @@
...
+
+declare(strict_types=1);
+
+namespace App\Policies;
+
+use App\Domain\Auth\Models\UserModel;
+use Illuminate\Auth\Access\HandlesAuthorization;
+use Spatie\Activitylog\Models\Activity;
+
+final class ActivityPolicy
```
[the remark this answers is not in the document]
LucWollants: To be more precise `App/Domain/Activity/Policies` or `App/Domain/ActivityLog/Policies`

PR #74, opened 2023-01-10
PPF-109 Do not restore contacts

steftrenson:
### Added
- Added a `ContactPolicy` so it's not possible to restore deleted contacts.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-109

### DIFF
--- a/app/Policies/ContactPolicy.php
+++ b/app/Policies/ContactPolicy.php
@@ -1,46 +1,48 @@
 <?php
 
+declare(strict_types=1);
+
 namespace App\Policies;
 
 use App\Domain\Auth\Models\UserModel;
 use App\Domain\Contacts\Models\ContactModel;
 use Illuminate\Auth\Access\HandlesAuthorization;
 
-class ContactPolicy
+final class ContactPolicy
 {
     use HandlesAuthorization;
 
-    public function viewAny(UserModel $userModel)
+    public function viewAny(UserModel $userModel): bool
     {
         return true;
     }
 
-    public function view(UserModel $userModel, ContactModel $contactModel)
+    public function view(UserModel $userModel, ContactModel $contactModel): bool
     {
         return true;
     }
 
-    public function create(UserModel $userModel)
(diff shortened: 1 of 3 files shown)


app/Policies/ContactPolicy.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return true;
+    }
+
+    public function delete(UserModel $userModel, ContactModel $contactModel): bool
+    {
+        return true;
+    }
+
+    public function restore(UserModel $userModel, ContactModel $contactModel): bool
```
JonasVHG, line 40: You can still `replicate()` a deleted contact. This will give a unique contraint eror when trying to save it, so it isn't a major issues.
But to me it looks more clean to also block replication, if you block restoring.

bertramakers: I was wondering, how do you know exactly which methods you can implement on the policy? I noticed that these policies do not use interfaces. Is there a list of possible methods somewhere? I checked the docs but it seems there is no complete list documented there (there is no mention of `replicate` for example) https://laravel.com/docs/9.x/authorization#creating-policies

app/Policies/ContactPolicy.php
```diff
@@ -0,0 +1,49 @@
...
+
+declare(strict_types=1);
+
+namespace App\Policies;
+
+use App\Domain\Auth\Models\UserModel;
+use App\Domain\Contacts\Models\ContactModel;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+final class ContactPolicy
```
LucWollants, line 11: I would store this class inside the `Domain` namespace, inside a sub folder `Policies`

LucWollants (approved):

app/Policies/ContactPolicy.php
JonasVHG: For a Nova specific list: https://nova.laravel.com/docs/4.0/resources/authorization.html#undefined-policy-methods

PR #76, opened 2023-01-13
PPF-117 unlink contact in Inisghtly

steftrenson:
### Added
- `unlinkContact` on `OpportunityResource`.

### Changed
- Inisghtly listener to delete a contact now unlinks a contact.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-117

### DIFF
--- a/app/Insightly/Resources/InsightlyOpportunityResource.php
+++ b/app/Insightly/Resources/InsightlyOpportunityResource.php
@@ -6,6 +6,7 @@ namespace App\Insightly\Resources;
 
 use App\Domain\Contacts\ContactType;
 use App\Domain\Integrations\Integration;
+use App\Insightly\Exceptions\ContactCannotBeUnlinked;
 use App\Insightly\InsightlyClient;
 use App\Insightly\Objects\OpportunityStage;
 use App\Insightly\Serializers\LinkSerializer;
@@ -83,6 +84,9 @@ final class InsightlyOpportunityResource implements OpportunityResource
         $this->insightlyClient->sendRequest($request);
     }
 
+    /**
+     * @throws ContactCannotBeUnlinked
+     */
     public function unlinkContact(int $opportunityId, int $contactId): void
     {
         $getLinksRequest = new Request(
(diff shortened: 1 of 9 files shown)


app/Insightly/Resources/InsightlyOpportunityResource.php
```diff
@@ -82,4 +83,40 @@ public function linkContact(int $opportunityId, int $contactId, ContactType $con
 
         $this->insightlyClient->sendRequest($request);
     }
+
+    /**
+     * @throws ContactCannotBeUnlinked
+     */
+    public function unlinkContact(int $opportunityId, int $contactId): void
+    {
+        $getLinksRequest = new Request(
```
LucWollants, line 92: This can become a reusable method on the opportunities resource. Because later on, we will also need to have it for projects, depending on the state/stage of an Integration.

app/Insightly/Resources/InsightlyOpportunityResource.php
```diff
@@ -82,4 +83,40 @@ public function linkContact(int $opportunityId, int $contactId, ContactType $con
...
+            $linkName = $opportunityLink['LINK_OBJECT_NAME'];
+            $linkObjectId = $opportunityLink['LINK_OBJECT_ID'];
+
+            if ($objectId === $opportunityId && $linkName === 'Contact' && $linkObjectId === $contactId) {
+                $linkId = $opportunityLink['LINK_ID'];
+            }
+        }
+
+        if ($linkId === null) {
+            throw new ContactCannotBeUnlinked();
```
LucWollants, line 112: Not sure we need to throw because that will result in a failed job that can never be fixed. Maybe just logging this as a warning is enough?

LucWollants (changes requested):
I think it can be useful to add method to retrieve links from Insightly for opportunities and projects

app/Insightly/Resources/InsightlyOpportunityResource.php
steftrenson: Good point. Will do!

app/Insightly/Resources/InsightlyOpportunityResource.php
steftrenson: I think we need to throw here, as we want to unlink it, but it can't be done. If we don't check the linkId the code will fail. But you're right it shouldn't make the job fail. I'll add a catch statement in the listener.

app/Insightly/Resources/InsightlyOpportunityResource.php
steftrenson: I'll make it a private function for now, as I don't know yet which properties should be exposed and don't want to expose everything yet. We can revisit when we need to.

LucWollants (approved):

PR #77, opened 2023-01-13
PPF-113 export coupons

JonasVHG:
### Fixed

---
Ticket: https://jira.uitdatabank.be/browse/PPF-113


app/Nova/Resources/Integration.php
```diff
@@ -143,6 +144,8 @@ public function fields(NovaRequest $request): array
 
             HasMany::make('Contacts'),
 
+            HasOne::make('Coupon'),
```
LucWollants, line 147: I would make this a [computed](https://nova.laravel.com/docs/1.0/resources/fields.html#computed-fields) field. It feels weird that the coupon field has its own panel and is not just included inside the general Integration panel.

PR #78, opened 2023-01-13
PPF-120 Coupon isDistributed

JonasVHG:
### Added
- Enum `CouponStatus`
- Added `isDistributed` to `Coupon`

---
Ticket: https://jira.uitdatabank.be/browse/PPF-120

### DIFF
--- a/app/Domain/Coupons/Coupon.php
+++ b/app/Domain/Coupons/Coupon.php
@@ -13,7 +13,6 @@ final class Coupon
         public readonly bool $isDistributed,
         public readonly ?UuidInterface $integrationId,
         public readonly string $code,
-        public readonly CouponStatus $status,
     ) {
     }
 }
--- a/app/Domain/Coupons/Coupon.php
+++ b/app/Domain/Coupons/Coupon.php
@@ -10,8 +10,10 @@ final class Coupon
 {
     public function __construct(
         public readonly UuidInterface $id,
+        public readonly bool $isDistributed,
         public readonly ?UuidInterface $integrationId,
-        public readonly string $code
+        public readonly string $code,
+        public readonly CouponStatus $status,
     ) {
     }
 }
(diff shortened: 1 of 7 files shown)


app/Domain/Coupons/Coupon.php
```diff
@@ -10,8 +10,10 @@ final class Coupon
 {
     public function __construct(
         public readonly UuidInterface $id,
+        public readonly bool $isDistributed,
         public readonly ?UuidInterface $integrationId,
-        public readonly string $code
+        public readonly string $code,
+        public readonly CouponStatus $status,
```
LucWollants, line 16: This field can be derived from `isDistributed` and `integrationId` so I would not make it a constructor argument. If you make it an argument this could lead to wrong combinations.

database/migrations/2023_01_13_142805_update_coupons_table.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('coupons', static function (Blueprint $table) {
+            $table->boolean('is_distributed')->after('id')->default(false);
```
LucWollants, line 13: It seems more logical to make this just the last column?

LucWollants (changes requested):
I would refactor this a bit to make it a derived or computed field.

database/migrations/2023_01_13_142805_update_coupons_table.php
JonasVHG: To me it seemed logical to put it next to `integration_id` because they belong together to calculate the status.
But regardless it doesn't change the working of the application so putting new columns at the end of a table might also be preferable.

LucWollants (approved):

PR #89, opened 2023-01-16
PPF-90 Link existing Insightly contact on create

steftrenson:
### Added
- Search function on contacts resource.

### Changed
- Creating a contact will check first if there is an existing contact in Insightly with the same email and link that.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-90


app/Insightly/Resources/InsightlyContactResource.php
```diff
@@ -55,4 +56,21 @@ public function delete(int $id): void
 
         $this->insightlyClient->sendRequest($request);
     }
+
+    /**
+     * @return int[]
+     */
+    public function findByEmail(string $email): array
```
LucWollants, line 63: If we return only the `id` of a contact I think we need to make that more clear in the method name. Maybe something like `findContactIdsByEmail` or `findIdsByEmail`?

app/Insightly/Resources/InsightlyContactResource.php
```diff
@@ -55,4 +56,21 @@ public function delete(int $id): void
...
+    {
+        $request = new Request(
+            'GET',
+            $this->path . "Search?field_name=email_address&field_value=$email&brief=true"
+        );
```
simon-debruijn, line 68:
```suggestion
        $request = new Request(
            'GET',
            $this->path . "Search",
            [
                'query' => [
                    'field_name' => "email_address",
                    'field_value' => $email,
                    'brief' => true,
                 ]
            ]
        );
```

If I read the GuzzleHttp docs correctly this could be an option to make the code a bit more readable 🙂

simon-debruijn (approved):

app/Insightly/Resources/InsightlyContactResource.php
bertramakers: Yes but no :) This particular class has `$headers` as the third argument, not `$options` that you're likely thinking of that you can use on `GuzzleHttp\Client::request()`.

See the `use GuzzleHttp\Psr7\Request` statement above and the documentation: https://docs.guzzlephp.org/en/stable/psr7.html (or jump to the class definition in PHPStorm)

It's a good idea of course but then the code would need to be changed to use `GuzzleHttp\Client::request()`.

bertramakers: I think the reason for this difference between `GuzzleHttp\Psr7\Request` (a class that represents a request, without sending it yet) and `GuzzleHttp\Client::request()` (a method that immediately sends a request with the given arguments) is that `GuzzleHttp\Psr7\Request` implements the PSR-7 `RequestInterface` (https://www.php-fig.org/psr/psr-7/) which does not support all of those options that Guzzle supports. So instead if you use that class you need to pass them when calling `GuzzleHttp\Client::sendRequest()` (as the second request), or you do not use that class at all and just call `GuzzleHttp\Client::request()`.

In the case of query parameters, the approach in the PSR-7 standard is that the PSR-7 request contains an instance of the PSR-7 UriInterface and that has methods to add/edit/remove query parameters. But the interfaces do not define the constructor method of the implementation, and the Guzzle team chose to accept a string as a URI and they convert it to an URI object under the hood. And in that case the query parameters have to be included in the string that represents the complete URI.

I think the reason for this discrepancy is that Guzzle existed before the PSR-7 standards, so they started off with a client class that had methods to send requests and those methods accepted some "options". And when the PSR-7 standard was introduced and adopted, they shoehorned it in in a way that works but is a bit confusing IMHO.

simon-debruijn: Too bad you can't pass those same options to the request class 🙁 Thanks for the in-depth explaination 😉

bertramakers: While your exact suggestion is not technically possible, we could put the `$options` in the `sendRequest()` below

JonasVHG (approved):

app/Insightly/Resources/InsightlyContactResource.php
LucWollants: I wonder how we can put the `$options` in the `sendRequest`? Because the [PSR-18 ClientInterface](https://github.com/php-fig/http-client/blob/master/src/ClientInterface.php) has only one parameter which is the `RequestInterface`.

LucWollants: But I think we could use the built-in PHP method [http_build_query](https://www.php.net/manual/en/function.http-build-query.php)

LucWollants: If needed we call easily refactor this in a follow-up pull request.

bertramakers: I thought we were using Guzzle since that was what the original comment was about, not the PSR-18 ClientInterface. The Guzzle client has a `send()` method (not `sendRequest()`, my bad) that accepts the `$options` as a second argument after the PSR-7 `Request` object. https://github.com/guzzle/guzzle/blob/master/src/ClientInterface.php#L30

If you're using the PSR-18 ClientInterface you cannot use the `$options` of course because the implementation won't necessarily be Guzzle so the PSR-18 standard doesn't support those Guzzle-specific options.

> But I think we could use the built-in PHP method [http_build_query](https://www.php.net/manual/en/function.http-build-query.php)

That works, a PSR-7 alternative is:
```
$request = new Request(
  'GET',
  (new Uri($this->path . 'Search')) // 'use GuzzleHttp\Psr7\Uri'
    ->withQueryParams(
      [
        'field_name' => 'email_address',
        'field_value' => $email,
        'brief' => true,
      ]
    )
);

PR #95, opened 2023-01-19
PPF-124 Migrate project aanvraag data

LucWollants:
### Added
- Added migration script for coupons (this is a prerequisite for projects migration)
- Added migration script for projects
  - Project data gets manipulated to an integration before saving
  - Coupon gets linked to integration
  - Insightly mapping gets created for the new integration to the existing Insightly project or opportunity
  - A contact gets created from Insightly data (as a fallback from UiTiD data with email and nick)
  - UiTiD consumers get created (missing UiTiD consumers will be handled in separate script, just like the missing Auth0 clients)
- Added exports from Project Aanvraag (this gives better visibility of what will be imported and a deploy makes sure the files are present)
  - `projects.csv` exported CSV data from `project` table
  - `coupons`csv` exported CSV data from `coupon` table

### Notes
- When doing a code review it will be most efficient to review the final result and not the individual commits. The migration was implemented with trial and error, so most of the commits are no longer meaningful.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-124


steftrenson (approved):
LGTM

app/Console/Migrations/MigrateProjects.php
```diff
@@ -0,0 +1,380 @@
...
+            []
+        );
+        $this->integrationRepository->save($integration);
+
+        IntegrationModel::query()->where('id', '=', $integrationId)->update([
+            'migrated_at' => Carbon::now(),
+        ]);
+    }
+
+    private function migrateCoupon(UuidInterface $integrationId, string $couponCode): bool
```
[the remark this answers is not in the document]
LucWollants: I removed all return statements.

app/Console/Migrations/MigrateProjects.php
```diff
@@ -0,0 +1,380 @@
...
+        if ($projectAsArray[6] === (string) config('insightly.integration_types.search_api')) {
+            $integrationType = IntegrationType::SearchApi;
+        }
+
+        $integration = new Integration(
+            $integrationId,
+            $integrationType,
+            $projectAsArray[3],
+            $projectAsArray[16] !== 'NULL' ? $projectAsArray[16] : '',
+            Uuid::fromString('b46745a1-feb5-45fd-8fa9-8e3ef25aac26'), // TODO: should be correct subscription plan
```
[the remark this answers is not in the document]
LucWollants: I created a constant for the UUID of the basic plan. This will need some extra finetuning and input from Corneel before we have a final solution.

app/Console/Migrations/MigrateProjects.php
```diff
@@ -0,0 +1,380 @@
...
+            $this->info('---');
+        }
+
+        return 0;
+    }
+
+    private function migrateIntegration(UuidInterface $integrationId, array $projectAsArray): void
+    {
+        $status = IntegrationStatus::Draft;
+        if ($projectAsArray[7] === 'active') {
```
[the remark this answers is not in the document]
LucWollants: I created a wrapper class `MigrationProject`

app/Console/Migrations/MigrateProjects.php
```diff
@@ -0,0 +1,340 @@
...
+        $response = $this->oauthClientProduction->request(
+            'GET',
+            'user/search?userId=' . $uitId,
+            ['http_errors' => false]
+        );
+        $status = $response->getStatusCode();
+
+        if ($status !== 200) {
+            $this->warn($uitId . ' - did not find user with UiTiD ' . $uitId);
+            return null;
```
simon-debruijn, line 250: You could make this less specific,

Preferably something like

`if (!$status->ok)`

I don't know the api

or

`if ($status < 200 || $status > 299)`

app/Console/Migrations/MigrateProjects.php
```diff
@@ -0,0 +1,340 @@
...
+            'GET',
+            'serviceconsumer/apikey/' . $apiKey,
+            ['http_errors' => false]
+        );
+
+        $status = $response->getStatusCode();
+        if ($status !== 200) {
+            $this->warn($integrationId . ' - did not find UiTiD consumer with API key ' . $apiKey);
+            return null;
+        }
```
simon-debruijn, line 322: Same comment

simon-debruijn (approved):

app/Console/Migrations/MigrateProjects.php
LucWollants: I wanted to make it very strict. I tested this and only status code `200` is a valid response.

simon-debruijn: Ok fair enough 😉

bertramakers: In general I do agree with @simon-debruijn that any 2xx response should be fine in theory, but I also agree with @LucWollants that in this case 200 is the only actual correct response. Because for example `201 Created` or `204 No Content` (other codes that I would accept in other cases) do not make sense for this `GET` request. `200 OK` is the only one that could return the data that we need. (See also https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses)

simon-debruijn: Makes sense indeed 👍🏻

PR #97, opened 2023-01-20
PPF-123 Refactor InsightlyMapping

JonasVHG:
### Changed
- Refactored update migration scripts into the create migration script: `add_status_to_integration`, `add_invoice_emial_to` & `update_coupons`
- Refactored migration script for `insightly_mappings` to make `id` non-unique
- Don't use primary key in `InsightlyMappingModel`
- use `firstOrFail()` for the moment, when using `getById()` in `InsightlyMappingRepository`

---
Ticket: https://jira.uitdatabank.be/browse/PPF-123


app/Insightly/Repositories/EloquentInsightlyMappingRepository.php
```diff
@@ -24,7 +24,7 @@ public function save(InsightlyMapping $insightlyMapping): void
     public function getById(UuidInterface $id): InsightlyMapping
     {
         /** @var InsightlyMappingModel $insightlyMappingModel */
-        $insightlyMappingModel = InsightlyMappingModel::query()->findOrFail($id);
+        $insightlyMappingModel = InsightlyMappingModel::query()->where('id', $id)->firstOrFail();
```
JonasVHG, line 27: This will have to be refactored when we have both Projects & Opportunities in `insightly_mappings`.

LucWollants: I think we will have to add a `getByIdAndType` method and remove `getById`. But that can be done in a later pull request.

simon-debruijn (approved):

app/Insightly/Models/InsightlyMappingModel.php
```diff
@@ -12,6 +12,8 @@ final class InsightlyMappingModel extends UuidModel
     use SoftDeletes;
 
     protected $table = 'insightly_mappings';
+    protected $primaryKey = null;
```
LucWollants, line 15: I think this class should now extend from `Model` instead of `UuidModel`

LucWollants: `Auth0ClientModel` is a good example.

LucWollants (approved):

PR #111, opened 2023-01-23
PPF-125 Activate Integration with Coupon

JonasVHG:
### Added

- method `useOnIntegration()` on `CouponModel`
- method `activeWithCoupon()` on `IntegrationModel`
- Event `IntegrationActivatedWithCoupon` & Test & Listener `ActivateIntegrationWithCoupon`
- method `opportunityToLink()` on `LinkSerializer`
- class `CouponSerializer` 
- method `updateState()` on `InsightlyOpportunityResource`
- methods `updateWithCoupon()`, `linkOpportunity()` & `linkContact()` on `InsightlyProjectResource`
- Creation of an Insightly Project when activating an `Integration` with a Coupon

### Changed
- Refactored `activateWithCouponCode()` on `IntegrationRepository`

### Removed
- `INSIGHTLY_API_KEY` from `phpunit.xml`

### Still TODO in following PR
- Linking contacts to the `Project`
- Setting the correct `Stage` & `State` on a Project

---
Ticket: https://jira.uitdatabank.be/browse/PPF-125

### DIFF
--- a/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
+++ b/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
@@ -248,7 +248,6 @@ final class EloquentIntegrationRepositoryTest extends TestCase
 
         $this->integrationRepository->save($searchIntegration);
 
-        $this->withoutEvents();
         $this->integrationRepository->activateWithCouponCode($integrationId, $couponCode);
 
         $this->assertDatabaseHas('integrations', [
--- a/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
+++ b/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
@@ -248,6 +248,7 @@ final class EloquentIntegrationRepositoryTest extends TestCase
 
         $this->integrationRepository->save($searchIntegration);
 
+        $this->withoutEvents();
         $this->integrationRepository->activateWithCouponCode($integrationId, $couponCode);
 
         $this->assertDatabaseHas('integrations', [
--- a/tests/Insightly/Listeners/ActivateIntegrationWithCouponTest.php
+++ b/tests/Insightly/Listeners/ActivateIntegrationWithCouponTest.php
@@ -1,0 +1,172 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Insightly\Listeners;
+
+use App\Domain\Coupons\Coupon;
+use App\Domain\Coupons\Repositories\CouponRepository;
+use App\Domain\Integrations\Events\IntegrationActivatedWithCoupon;
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\IntegrationStatus;
+use App\Domain\Integrations\IntegrationType;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Insightly\InsightlyMapping;
+use App\Insightly\Listeners\ActivateIntegrationWithCoupon;
+use App\Insightly\Objects\OpportunityStage;
+use App\Insightly\Objects\OpportunityState;
+use App\Insightly\Repositories\InsightlyMappingRepository;
+use App\Insightly\Resources\ResourceType;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+use Tests\MockInsightlyClient;
+
+final class ActivateIntegrationWithCouponTest extends TestCase
+{
+    use MockInsightlyClient;
+
+    private ActivateIntegrationWithCoupon $listener;
+
+    private IntegrationRepository&MockObject $integrationRepository;
+
+    private InsightlyMappingRepository&MockObject $insightlyMappingRepository;
+
+    private CouponRepository&MockObject $couponRepository;
+
+    private UuidInterface $integrationId;
+
+    private string $couponCode;
+
+    protected function setUp(): void
+    {
+        $this->integrationId = Uuid::uuid4();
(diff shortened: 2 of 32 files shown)


tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
```diff
@@ -248,6 +248,7 @@ public function test_it_can_activate_with_a_coupon(): void
 
         $this->integrationRepository->save($searchIntegration);
 
+        $this->withoutEvents();
```
LucWollants, line 251: Can this not be solved here https://github.com/cultuurnet/publiq-platform/blob/bc34e139efbafb81de7e192b5868518fb1511063/tests/TestCase.php#L19-L21 ?

bertramakers: FYI I think you can just do `Model::unsetEventDispatcher()` once because it's a static method defined on `Model`

JonasVHG: The tests complain about it: `No query results for model [App\Insightly\Models\InsightlyMappingModel` Though maybe I could look to fix it with a `Mock`

LucWollants: I think you should not use `Model::unsetEventDispatcher();` but `IntegrationModel::unsetEventDispatcher();`

LucWollants: I can even run the tests locally when I remove the line with `$this->withoutEvents();` 🤔

bertramakers: `Model::unsetEventDispatcher()` disables it for all models. See https://github.com/laravel/framework/issues/29727
So that should be sufficient to replace those 3 separate calls to `unsetEventDispatcher()`.
But that comment was just about those 3 specific existing lines, not about your question about `withoutEvents()`.

tests/Insightly/Listeners/ActivateIntegrationWithCouponTest.php
```diff
@@ -0,0 +1,172 @@
...
+            ->willReturn($integration);
+
+        return $integration;
+    }
+
+    private function givenThereIsAnOpportunityMapping(UuidInterface $integrationId): InsightlyMapping
+    {
+        $insightlyMapping = new InsightlyMapping(
+            $integrationId,
+            42,
```
LucWollants, line 144: Maybe not bad to make this Insightly id a parameter. Later we should also extract these given methods to a reusable component like a class or trait.

tests/Insightly/Listeners/ActivateIntegrationWithCouponTest.php
```diff
@@ -0,0 +1,172 @@
...
+    public function test_it_creates_a_project_when_activating_with_a_coupon(): void
+    {
+        // Given
+        $insightlyOpportunityId = 42;
+        $insightlyProjectId = 51;
+        $integration = $this->givenThereIsAnIntegrationWithId($this->integrationId);
+        $this->givenThereIsAnOpportunityMapping($this->integrationId);
+        $coupon = $this->givenThereIsACoupon($this->integrationId);
+
+        $this->integrationRepository->activateWithCouponCode($this->integrationId, $this->couponCode);
```
LucWollants, line 71: There is no need to call this method `activateWithCouponCode` on a mock.

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
LucWollants: The issue is something totally different. In the base class `TestCase` we did put the `unsetEventDispatcher` because we wanted to prevent the dispatch from being called after "normal" Eloquent model interaction.

But now we fire do `dispatch` inside a custom method so that will not work https://github.com/cultuurnet/publiq-platform/blob/4d3647afacf18175babfd03d259e42f629eeb832/app/Domain/Integrations/Models/IntegrationModel.php#L56

LucWollants: The solution is to use the trait `WithoutEvents`. This trait will be executed inside the `setUp` method.

The disabling of events has to happen exactly at that code location because we have 2 distinct ways of using the `dispatch`. The first and common approach is inside the static `booted` method and the new approach is inside the `activeWithCoupon` method.

If we disable the events later, for example in our custom `setUp` method then the `dispatch` inside the `booted` method are still done.

tests/Insightly/Listeners/ActivateIntegrationWithCouponTest.php
LucWollants: Removed in https://github.com/cultuurnet/publiq-platform/pull/111/commits/a60d00a37afe05f53dbbca1607d9abfbe8cb36ab

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
LucWollants: Solved in https://github.com/cultuurnet/publiq-platform/pull/111/commits/61e5e015ee961605d7c397ed7b656508e41db81a

tests/Insightly/Listeners/ActivateIntegrationWithCouponTest.php
LucWollants: Solved in https://github.com/cultuurnet/publiq-platform/pull/111/commits/ea604ce6c45cfc2c07f438be60298cbd1fe6011d

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
bertramakers: Magic 🪄

LucWollants (approved):

PR #114, opened 2023-01-26
PPF-118 Update contact email flow in Insightly

steftrenson:
### Changed
- Updating the email of a contact will result in linking a new contact in Insightly rather than updating the already linked contact.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-118

### DIFF
--- a/app/Insightly/InsightlyContactLink.php
+++ b/app/Insightly/InsightlyContactLink.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Insightly;
+
+use App\Domain\Contacts\Contact;
+
+final class InsightlyContactLink implements ContactLink
+{
+    public function link(Contact $contact): int
+    {
+        // TODO: Implement link() method.
+    }
+}
--- a/app/Insightly/InsightlyContactLink.php
+++ b/app/Insightly/InsightlyContactLink.php
@@ -5,11 +5,23 @@ declare(strict_types=1);
 namespace App\Insightly;
 
 use App\Domain\Contacts\Contact;
+use Illuminate\Support\Arr;
 
 final class InsightlyContactLink implements ContactLink
 {
+    public function __construct(private readonly InsightlyClient $insightlyClient)
+    {
+    }
+
     public function link(Contact $contact): int
     {
-        // TODO: Implement link() method.
+        $contactIds = $this->insightlyClient->contacts()->findIdsByEmail($contact->email);
+
+        if (empty($contactIds)) {
+            return $this->insightlyClient->contacts()->create($contact);
+        }
+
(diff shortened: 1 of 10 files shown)


app/Insightly/InsightlyContactLink.php
```diff
@@ -0,0 +1,27 @@
...
+use App\Domain\Contacts\Contact;
+use Illuminate\Support\Arr;
+
+final class InsightlyContactLink implements ContactLink
+{
+    public function __construct(private readonly InsightlyClient $insightlyClient)
+    {
+    }
+
+    public function link(Contact $contact): int
```
LucWollants, line 16: For me, it is not so clear what this method does. Maybe it is better for readability to just inline this inside the handlers?

app/Insightly/Listeners/UpdateContact.php
```diff
@@ -47,6 +54,48 @@ public function handle(ContactUpdated $contactUpdated): void
         );
     }
 
+    private function handleContactEmailChange(Contact $contact): void
```
LucWollants, line 57: I think when the email is changed other properties like first name and last name could also change. But this is not handled?

tests/Insightly/InsightlyContactLinkTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $this->contactLink = new InsightlyContactLink($this->insightlyClient);
+    }
+
+    public function test_it_creates_a_new_contact_when_the_email_is_unique(): void
+    {
+        $email = 'info@publiq.be';
+        $insightlyId = 42;
+
+        $contact = $this->buildContactWithEmail($email);
```
LucWollants, line 32: It seems that all the contacts have the same values inside this unit test, maybe it can become a property that gets created inside the `setUp` method?

app/Insightly/InsightlyContactLink.php
LucWollants: Maybe a better name is something like `getUniqueContact` which can become part of `ContactResource`?

LucWollants: Or even better modify `ContactResource:create` to enforce uniqueness

steftrenson: Thinking about it, it feels more like a `findOrCreateByEmail()` function. The hard thing is determining where we should guard the domain rule that Contacts in Insightly stay unique on email address.

app/Insightly/Listeners/UpdateContact.php
steftrenson: Should this also happen on the create then? If someone creates a contact with an email address that already exists in Insightly, should the name also be updated?

PR #115, opened 2023-01-26
PPF-125 correct stage state project

JonasVHG:
### Added
- `updateStage()` on `ActivateIntegrationWithCoupon`
- It sets the correct stage & state after for a `Project` after activating the `Integration`

### Changed
- use consistent naming for `insightlyId` in `Insightly` Interfaces (always `$id`).

### Fixed
- Insightly config for `ProjectStage`
- Don't use hardcodes path in Insightly calls 

---
Ticket: https://jira.uitdatabank.be/browse/PPF-125


app/Insightly/Resources/InsightlyProjectResource.php
```diff
@@ -106,11 +121,11 @@ public function linkContact(int $projectId, int $contactId, ContactType $contact
         $this->insightlyClient->sendRequest($request);
     }
 
-    private function get(int $id): array
+    private function get(int $projectId): array
     {
         $request = new Request(
             'GET',
-            'Projects/' . $id
+            'Projects/' . $projectId
```
LucWollants, line 128: I noticed we sometimes use the property `$this-path` and sometimes a hardcoded string. Best to make consistent use of the `$path` property. While checking this please also check the other resources.

tests/TestCase.php
```diff
@@ -16,8 +14,6 @@ abstract class TestCase extends BaseTestCase
     protected function setUp(): void
     {
         parent::setUp();
-        ContactModel::unsetEventDispatcher();
-        IntegrationModel::unsetEventDispatcher();
-        OrganizationModel::unsetEventDispatcher();
+        Model::unsetEventDispatcher();
```
LucWollants, line 17: To fix this you will need to merge the `main` branch.

app/Insightly/Resources/ProjectResource.php
```diff
@@ -7,16 +7,19 @@
...
 use App\Domain\Integrations\Integration;
 use App\Insightly\Objects\ProjectStage;
+use App\Insightly\Objects\ProjectState;
 
 interface ProjectResource
 {
     public function create(Integration $integration): int;
 
-    public function updateWithCoupon(int $insightlyId, string $couponCode): void;
+    public function updateWithCoupon(int $projectId, string $couponCode): void;
```
LucWollants, line 16: I noticed we now have 4 flavors inside the resource interfaces for the Insightly id:
- `id`
- `insightlyId`
- `projectId`

I would pick one and make it consistent over all the interfaces.

LucWollants: How will we handle `ContactResource` and `OrganizationResource`?

I think that using `id` is actually just fine. We are in the context of Insightly and all the other interfaces like the repositories use `id` for the object that is manipulated. For example `ContactRepository::getById(UuidInterface $id)`

LucWollants (approved):

PR #132, opened 2023-02-01
Refactored save contact to either update or create

LucWollants:
### Changed
- Refactored saving a contact with the `ContactRepository` to either update a contact or create a new contact based on the id

### Note
- As preparation for the migration
- In the future more `save` methods will need to be refactored to do a `create` or `update`

### DIFF
--- a/app/Domain/Contacts/Repositories/EloquentContactRepository.php
+++ b/app/Domain/Contacts/Repositories/EloquentContactRepository.php
@@ -13,14 +13,19 @@ final class EloquentContactRepository implements ContactRepository
 {
     public function save(Contact $contact): void
     {
-        ContactModel::query()->create([
-            'id' => $contact->id->toString(),
-            'integration_id' => $contact->integrationId->toString(),
-            'type' => $contact->type->value,
-            'first_name' => $contact->firstName,
-            'last_name' => $contact->lastName,
-            'email' => $contact->email,
-        ]);
+        ContactModel::query()->updateOrCreate(
+            [
+                'id' => $contact->id->toString(),
+            ],
+            [
+                'id' => $contact->id->toString(),
+                'integration_id' => $contact->integrationId->toString(),
+                'type' => $contact->type->value,
+                'first_name' => $contact->firstName,
+                'last_name' => $contact->lastName,
+                'email' => $contact->email,
+            ]
+        );
     }
 
     public function getById(UuidInterface $id): Contact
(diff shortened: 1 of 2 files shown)


app/Domain/Contacts/Repositories/EloquentContactRepository.php
```diff
@@ -13,14 +13,19 @@ final class EloquentContactRepository implements ContactRepository
...
-            'first_name' => $contact->firstName,
-            'last_name' => $contact->lastName,
-            'email' => $contact->email,
-        ]);
+        ContactModel::query()->updateOrCreate(
+            [
+                'id' => $contact->id->toString(),
+            ],
+            [
+                'id' => $contact->id->toString(),
```
steftrenson, line 21: Is this still needed here? Because a new contact won't have an id, right?

LucWollants: I think it is needed because a new contact has an `id` which is part of the domain object.
This `id` is a UUID and is created and set before saving the contact to the database.

bertramakers: If the contact already has a UUID before it's saved to the DB, you need an `updateOrCreate()` an explicitly insert/update the `id` column yes. Another approach would be to not give new contacts a UUID in the domain model and check here if a create or update needs to be done based on the UUID being null. But `null` values are always annoying and the app should be able to generate its own UUIDs without issues instead of needing to rely on the ORM to do that, so I also slightly prefer this approach.

tests/Domain/Contacts/Repositories/EloquentContactRepositoryTest.php
```diff
@@ -23,7 +23,7 @@ protected function setUp(): void
         $this->contactRepository = new EloquentContactRepository();
     }
 
-    public function test_it_can_save_a_contact(): void
+    public function test_it_can_create_a_contact(): void
```
bertramakers, line 26:
```suggestion
    public function test_it_can_create_and_update_a_contact(): void
```

bertramakers (approved):
My only suggestion is very minor, and everything else looks great, so already approving this 👌

app/Domain/Contacts/Repositories/EloquentContactRepository.php
steftrenson: Oh yeah, didn't fully realize it's a UUID. Carry on.

steftrenson (approved):

PR #133, opened 2023-02-01
PPF-118 Sync contact

steftrenson:
Splitting up the sync functionality with Insightly into a create, and an update listener was splitting up responsibilities too much. All logic on what needs to happen when you create or update a contact and how that should be reflected in Insightly is now available in SyncContact listener.

### Added
- `SyncContact` listener, which handles all business logic when a platform contact is created or updated and syncs it to Insightly.

### Removed
- `CreateContact` and `UpdateContact` listeners for Inightly.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-118


app/Insightly/Listeners/SyncContact.php
```diff
@@ -0,0 +1,127 @@
...
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
+    public function handleContactCreated(ContactCreated $contactCreated): void
+    {
+        $contact = $this->contactRepository->getById($contactCreated->id);
+        if (!SyncIsAllowed::forContact($contact)) {
+            return;
+        }
```
simon-debruijn, line 39:
```suggestion
        $contact = $this->contactRepository->getById($contactCreated->id);
        
        if (!SyncIsAllowed::forContact($contact)) {
            return;
        }
```

JonasVHG (approved):

app/Insightly/Listeners/SyncContact.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private function storeAndLinkContactAtInsightly(Contact $contact): void
+    {
+        $contactIds = $this->insightlyClient->contacts()->findIdsByEmail($contact->email);
+
+        if (empty($contactIds)) {
+            $contactInsightlyId = $this->insightlyClient->contacts()->create($contact);
+        } else {
+            $contactIds = array_values(Arr::sort($contactIds));
```
LucWollants, line 59: I vaguely remember that we discussed with @corneelwille to take the contact with the most linked projects and opportunities.

app/Insightly/Listeners/SyncContact.php
```diff
@@ -0,0 +1,127 @@
...
+use App\Insightly\InsightlyMapping;
+use App\Insightly\Repositories\InsightlyMappingRepository;
+use App\Insightly\Resources\ResourceType;
+use App\Insightly\SyncIsAllowed;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Support\Arr;
+use Psr\Log\LoggerInterface;
+
+final class SyncContact implements ShouldQueue
```
LucWollants, line 22: Should we also implement the `failed` method to get a nice processable error when it fails?

app/Insightly/Listeners/SyncContact.php
```diff
@@ -0,0 +1,127 @@
...
+            $this->insightlyClient->contacts()->update($contact, $contactInsightlyId);
+        }
+
+        $this->insightlyMappingRepository->save(new InsightlyMapping(
+            $contact->id,
+            $contactInsightlyId,
+            ResourceType::Contact
+        ));
+
+        $integrationMapping = $this->insightlyMappingRepository->getByIdAndType($contact->integrationId, ResourceType::Opportunity);
```
LucWollants, line 71: With a recent change to the mapping table, I think we also need to make sure to update the project mapping if it exists.

Do we also need to make sure the mapping exists in the database?

tests/Insightly/Listeners/SyncContactTest.php
```diff
@@ -0,0 +1,332 @@
...
+    {
+        $this->givenThereIsAContactForAnIntegration(ContactType::Contributor);
+
+        $this->insightlyClient->expects($this->never())
+            ->method('contacts');
+
+        $this->syncContact->handleContactUpdated(new ContactUpdated($this->contactId, true));
+    }
+
+    public function test_it_links_a_new_insightly_contact_when_platform_contact_was_created(): void
```
LucWollants, line 78: What is `platform` here?

steftrenson: `publiq platform`. Should I name it that explicit?

LucWollants: Or just remove it?

app/Insightly/Listeners/SyncContact.php
steftrenson: I don't really get why the opportunity mapping should be updated. The contact is linked or unlinked from an opportunity, but the opportunity itself doesn't change.

app/Insightly/Listeners/SyncContact.php
steftrenson: Done.

app/Insightly/Listeners/SyncContact.php
LucWollants: What I meant was that an integration can now have an opportunity and a project inside Insightly. So now the contact is only linked to the opportunity but it should also be linked to the project.

So that is also why it is important to validate that the mapping exists, an integration could have an opportunity and/or project mapping. (When doing the migration of the old Project Aanvraag data it will even only have a project mapping)

LucWollants (approved):

PR #162, opened 2023-02-14
PPF-139 refactor coupon resource

JonasVHG:
### Changed

- Hide `ID` from listview for Resource `Coupon`
- Reorder fields for Resource `Coupon` from: `Distributed`|`Integration`|`Coupon code` to `Coupon code`|`Distributed`|`Integration`
- Make all fields for Resource `Coupon` sortable in listView
- Add Filter for field `Distributed` on Resource `Coupon`
- Reintroduced Abstract class `Resource` to put sorting logic
- Default Sorting on `Coupon`, undistributed `Coupons` appear first

---
Ticket: https://jira.uitdatabank.be/browse/PPF-139


app/Nova/Resources/Coupon.php
```diff
@@ -32,20 +33,41 @@ final class Coupon extends Resource
         'is_distributed',
     ];
 
+    public static array $defaultSort = [
+        'is_distributed' => 'asc',
+    ];
+
+    public static function indexQuery(NovaRequest $request, $query): Builder
```
LucWollants, line 40: Keep in mind that the proposed solution https://github.com/laravel/nova-issues/issues/156#issuecomment-415974405 adds this `indexQuery` method inside the custom `Resource`. By doing this it works on every Resource. But we deleted this custom `Resource` because it had no custom methods. Maybe it's worth bringing that file back?

JonasVHG: As of now we don't need a default sort in the other `Resources`, but it is indeed prudent to assume this might change in the future. I'll make the changes.

app/Nova/Resource.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace App\Nova;
+
+use Illuminate\Database\Eloquent\Builder;
+use Laravel\Nova\Http\Requests\NovaRequest;
+use Laravel\Nova\Resource as NovaResource;
+
+abstract class Resource extends NovaResource
+{
+    public static ?array $defaultSort = null;
```
LucWollants, line 13: I think it's safer to make this protected. No need to set it from outside a Resource.
But I think `indexQuery` needs to stay public if I see how this works inside the Laravel Nova code?

LucWollants (approved):

app/Nova/Resource.php
JonasVHG: Good find, I'll make it protected.

PR #163, opened 2023-02-14
PPF-135 Create auth0 clients for migrated projects

LucWollants:
### Added
- Created command `CreateAuth0Clients` to create Auth0 clients for migrated integrations. If no saved Auth0 clients are found new ones are created on all environments.

---

Ticket: https://jira.uitdatabank.be/browse/PPF-135

### DIFF
--- a/app/Console/Commands/Migrations/CreateAuth0Clients.php
+++ b/app/Console/Commands/Migrations/CreateAuth0Clients.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Console\Migrations;
+namespace App\Console\Commands\Migrations;
 
 use App\Auth0\Auth0ClusterSDK;
 use App\Auth0\Repositories\Auth0ClientRepository;
--- a/app/Console/Commands/Migrations/MigrateCoupons.php
+++ b/app/Console/Commands/Migrations/MigrateCoupons.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Console\Migrations;
+namespace App\Console\Commands\Migrations;
 
 use App\Domain\Auth\Models\UserModel;
 use App\Domain\Coupons\Coupon;
--- a/app/Console/Commands/Migrations/MigrateProjects.php
+++ b/app/Console/Commands/Migrations/MigrateProjects.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Console\Migrations;
+namespace App\Console\Commands\Migrations;
 
 use App\Domain\Auth\Models\UserModel;
 use App\Domain\Contacts\Contact;
--- a/app/Console/Commands/Migrations/MigrationProject.php
+++ b/app/Console/Commands/Migrations/MigrationProject.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Console\Migrations;
+namespace App\Console\Commands\Migrations;
 
 use App\Domain\Integrations\IntegrationStatus;
 use App\Domain\Integrations\IntegrationType;
--- a/app/Console/Commands/Migrations/ReadCsvFile.php
+++ b/app/Console/Commands/Migrations/ReadCsvFile.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Console\Migrations;
+namespace App\Console\Commands\Migrations;
 
 trait ReadCsvFile
 {
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -4,18 +4,11 @@ declare(strict_types=1);
 
 namespace App\Console;
 
-use App\Console\Migrations\MigrateCoupons;
-use App\Console\Migrations\MigrateProjects;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
 
 final class Kernel extends ConsoleKernel
 {
-    protected $commands = [
-        MigrateCoupons::class,
-        MigrateProjects::class,
-    ];
-
     protected function schedule(Schedule $schedule): void
     {
     }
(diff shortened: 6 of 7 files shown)


app/Console/Commands/Migrations/CreateAuth0Clients.php
```diff
@@ -0,0 +1,80 @@
...
+use App\Auth0\Auth0ClusterSDK;
+use App\Auth0\Repositories\Auth0ClientRepository;
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\Models\IntegrationModel;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+final class CreateAuth0Clients extends Command
+{
+    protected $signature = 'auth0:create-auth0-clients';
```
bertramakers, line 16:
```suggestion
    protected $signature = 'auth0:create-missing-auth0-clients';
```

app/Console/Commands/Migrations/CreateAuth0Clients.php
```diff
@@ -0,0 +1,80 @@
...
+
+    protected $description = 'Create Auth0 clients for migrated integrations';
```
bertramakers, line 18:
```suggestion
    protected $description = 'Create Auth0 client(s) for integrations that are missing one or more';
```

bertramakers: I wouldn't specifically focus on migrated projects here, because we could also use this later to fix integrations that did not get one or more Auth0 clients due to bugs/Auth0 being down/...
You'd also need to change the code to not only loop over migrated integrations of course.
Maybe you can query all integrations that do not have 3 auth0 clients for example?

app/Console/Commands/Migrations/CreateAuth0Clients.php
LucWollants: I also changed the filename.

app/Console/Commands/Migrations/CreateAuth0Clients.php
LucWollants: I agree we can make this command more generic and create clients for integrations that are missing them.

> Maybe you can query all integrations that do not have 3 auth0 clients for example?
Will this not lead to a crash when there is already a client for a certain environment, because `createClientsForIntegration` creates all of them again? Unless the code gets changed to only create it for the missing environments?

bertramakers: Yeah you're right. You could add a getter to `Auth0ClusterSDK` to get a tenant SDK for the specific tenant(s) that an integration has no client(s) for and then use those directly. Or add a new `createMissingClientsForIntegration()` to `Auth0ClusterSDK`. Or maybe change the existing code in `createClientsForIntegration()` if that doesn't break anything yes.

LucWollants: I will create a new call `createClientForIntegrationOnAuth0Tenant`, because:
- I think it's best to not expose `Auth0TenantSDK` 
- Make a call that is reusable in another context

app/Auth0/Auth0ClusterSDK.php
```diff
@@ -32,4 +32,11 @@ public function createClientsForIntegration(Integration $integration): array
             )
         );
     }
+
+    public function createClientsForIntegrationOnAuth0Tenant(
+        Integration $integration,
+        Auth0Tenant $auth0Tenant
+    ): Auth0Client {
+        return $this->auth0TenantSDKs[$auth0Tenant->value]->createClientForIntegration($integration);
```
bertramakers, line 40: Note that in theory it's possible that there is no SDK set for a given tenant, for example when the config is left blank for that tenant. Is it okay for the script to just crash on that, under the assumption that you should only run it with non-empty config values for Auth0 since you explicitly do it to create clients? And/or, is it okay for this more generic method to ignore this potential error and leave it to the caller to deal with that?

app/Console/Commands/Migrations/CreateMissingAuth0Clients.php
```diff
@@ -84,7 +84,7 @@ private function createMissingAuth0ClientsForIntegration(Integration $integratio
      */
     private function getMigratedIntegrations(): Collection
     {
-        $migratedIntegrationModels = IntegrationModel::query()->whereNotNull('migrated_at')->get();
+        $migratedIntegrationModels = IntegrationModel::query()->get();
```
bertramakers, line 87: Can we query here specifically to only get the integrations that do not have three auth0 clients? Is that not possible with Eloquent?

app/Auth0/Auth0ClusterSDK.php
LucWollants: I think it's best to throw an error then, something like `Auth0TenantNotConfigured`?

app/Console/Commands/Migrations/CreateMissingAuth0Clients.php
```diff
@@ -89,7 +89,7 @@ private function createMissingAuth0ClientsForIntegration(Integration $integratio
      */
     private function getIntegrationsWithMissingAuth0Clients(): Collection
     {
-        $migratedIntegrationModels = IntegrationModel::query()->has('auth0Clients', '<',  3)->get();
+        $migratedIntegrationModels = IntegrationModel::query()->has('auth0Clients', '<', 3)->get();
```
bertramakers, line 92: Maybe also use the `count(Auth0Tenant::cases())` here like here? https://github.com/cultuurnet/publiq-platform/pull/163/files#diff-d994cfcee8ed4cd920c17e6dd109c4deac9172d85f18a9788ca72a565e60180bR60

I don't see another one getting added in the foreseeable future, but who knows 🤷‍♂️

bertramakers (approved):
Everything looks good to me aside from the one minor suggestion

app/Console/Commands/Migrations/CreateMissingAuth0Clients.php
LucWollants: What did I do there 🙈

PR #165, opened 2023-02-14
PPF-141 refactor resource organization

JonasVHG:
### Changed

- Hide `ID`, `VAT`, `Street` & `Country` from listview for `Organization`
- Make filter for `Zip` on listview for `Organization`
- reorder fields on listview `Organization` from `Name|Invoice Email|Zip|City` to `Organization` from `Name|Invoice Email|City|Zip`

---
Ticket: https://jira.uitdatabank.be/browse/PPF-141


app/Nova/Resources/Organization.php
```diff
@@ -53,22 +54,26 @@ public function fields(NovaRequest $request): array
...
                 ->rules('required', 'max:255'),
 
-            Text::make('Zip')
+            Text::make('City')
                 ->sortable()
                 ->rules('required', 'max:255'),
 
-            Text::make('City')
+            Text::make('Zip')
+                ->filterable()
```
LucWollants, line 71: Is this required? Because just typing `3000` in the search field will also work?

LucWollants (approved):

app/Nova/Resources/Organization.php
JonasVHG: I'll ask Corneel. An edge case might be if an `Organisation` has a name that contains like `1000Handen VZW` but has a zip of `2360` or `Tennis 2000` that has a zip of `9308`

PR #180, opened 2023-02-20
PPF-97 Activate integration with organization

LucWollants:
### Added
- Added listener to create a project inside Insightly when integration gets activated with organization details

---
Ticket: https://jira.uitdatabank.be/browse/PPF-97

### DIFF
--- a/app/Insightly/InsightlyServiceProvider.php
+++ b/app/Insightly/InsightlyServiceProvider.php
@@ -8,6 +8,7 @@ use App\Domain\Contacts\Events\ContactCreated;
 use App\Domain\Contacts\Events\ContactDeleted;
 use App\Domain\Contacts\Events\ContactUpdated;
 use App\Domain\Integrations\Events\IntegrationActivatedWithCoupon;
+use App\Domain\Integrations\Events\IntegrationActivatedWithOrganization;
 use App\Domain\Integrations\Events\IntegrationCreated;
 use App\Domain\Organizations\Events\OrganizationCreated;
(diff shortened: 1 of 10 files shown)


JonasVHG (approved):

app/Insightly/Listeners/CreateProjectWithCoupon.php
```diff
@@ -25,6 +18,7 @@
 final class CreateProjectWithCoupon implements ShouldQueue
 {
     use Queueable;
+    use CreatesProject;
```
bertramakers, line 21: My only (very minor) suggestion would be to make this trait a `CreateProject` class instead, maybe even invokable, which you instantiate directly in the constructor of `CreateProjectWithCoupon` and `CreateProjectWithOrganization` and run in their `handle()` methods. So basically the same but a class instead of a trait. But that way, the dependencies of the `CreateProject` class are clearer because it has its own properties that need to be injected (can be done via the `CreateProjectWithCoupon`/`CreateProjectWithOrganization` constructors of course)

bertramakers (approved):

app/Insightly/Listeners/CreateProjectWithCoupon.php
LucWollants: Very good remark. I also struggled with this design question during development. I will create an extra class `CreateProject` that has the required properties. I could even inject this class inside `CreateProjectWithCoupon` and `CreateProjectWithOrganization`.

PR #184, opened 2023-02-23
PPF-142 add resource policies

LucWollants:
### Added
- Added policies for `Integration`, `Organization`, and `Subscription`

### Changed
- Disable replicate and force delete on `Contact`

### Removed
- Removed the unused trait `HandlesAuthorization`

---
Ticket: https://jira.uitdatabank.be/browse/PPF-142


bertramakers (approved):
I'm not sure if contacts should be deletable, since I believe it could result in an integration not having a technical or functional contact? But maybe I'm overlooking something or they really need to be deletable for some reason even with that risk.

LucWollants:
> I'm not sure if contacts should be deletable, since I believe it could result in an integration not having a technical or functional contact? But maybe I'm overlooking something or they really need to be deletable for some reason even with that risk.

I think I can make the delete policy smart and prevent deleting the "last contributor"

corneelwille:
> > I'm not sure if contacts should be deletable, since I believe it could result in an integration not having a technical or functional contact? But maybe I'm overlooking something or they really need to be deletable for some reason even with that risk.
> 
> I think I can make the delete policy smart and prevent deleting the "last contributor" of an integration. @corneelwille does this sound like a good improvement?

Deleting a contact from an integration should be possible, since it is possible that a certain contact is not the contact for an integration anymore; e.g. in the following cases
* **An integration moves from one vendor to another**: The integration was built by IT supplier X but is now supported by supplier Y (the technical contact for the integration should no longer be the developer of IT supplier X)
* **Personnel changes** The integration was built by developer "Jan" from IT supplier X. "Jan" does not work for IT supplier X anymore. His colleague "An" is now the technical contact 

However, we must ensure that there is always a functional and technical contact for each integration. So we should ensure that (in the frontend)
* you can always remove a **contributor**
* you can delete a **technical contact** when you also enter a new one (= replace the contact)
* you can delete a **functional contact** when you also enter a new one (= replace the contact)

LucWollants:
I created https://jira.uitdatabank.be/browse/PPF-155 as a follow-up for the integrator user stories inside the dedicated frontend UI. For the admin user stories, I think we are fine for now inside Laravel Nova.

bertramakers:
But why delete a technical and/or functional contact, instead of just editing their name and email address?

corneelwille:
When a contact changes, the existing contact info can be updated inside publiq platform, but _not_ in Insightly. So, in the case "Bert Ramakers" (email address "bert@publiq.be") needs to be replaced by "Jonas Verhaeghe" (email address "jonas@publiq.be), we want
* the contact Bert Ramakers to be removed as a contact from the integration in publiq platform
* the contact Bert Ramakers to be removed as a linked contact from the integration in Insightly
* the contact Jonas Verhaeghe to be added as a contact to the integration in publiq platform
* the contact Jonas Verhaeghe to be added as a linked contact to the integration in Insightly

What we don't want:
* the contact Bert Ramakers may not be deleted from Insightly
* the contact information of Jonas Verhaeghe may not overwrite the existing contact information of the contact Bert Ramakers in Insightly

In short:
* update in publiq platform: yes
* update in Insightly: no (remove and add new one)

How this is handled in the code is not my decision

bertramakers:
> What if there are 2 functional contacts and you want to remove one of those functional contacts without replacing it?

I don't think it should ever be possible to have 2 functional contacts? I can't find a concrete source at the moment, but I always understood that an integration can only have 1 functional contact, 1 technical contact, and later we agreed to make it possible to add 0..n extra contributors to make it possible for extra users to view/edit the integration.

corneelwille:
Yes, it was indeed the idea for an integration to have 1 functional contact and 1 technical contact (no less and no more). 
- **Terminology page**: https://confluence.uitdatabank.be/display/UPW/PPF+Terminology
> An integration always has 1 functional contact and 1 technical contact. The functional contact can be the same as the technical contact.
- **user screens**: integrator / applicant can only add one functional and one technical contact: https://miro.com/app/board/uXjVOPoMLzA=/

LucWollants:
Up-to-date specs inside this planned ticket https://jira.uitdatabank.be/browse/PPF-155

PR #197, opened 2023-03-03
PPF-149 Block UiTiD consumers

LucWollants:
### Added
- Added `BlockConsumers` listener inside UiTiD domain which listens to the `IntegrationBlocked` event

---
Ticket: https://jira.uitdatabank.be/browse/PPF-149

### DIFF
--- a/app/UiTiDv1/Listeners/BlockConsumers.php
+++ b/app/UiTiDv1/Listeners/BlockConsumers.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\UiTiDv1\Listeners;
+
+use App\Domain\Integrations\Events\IntegrationBlocked;
+use App\UiTiDv1\Repositories\UiTiDv1ConsumerRepository;
+use App\UiTiDv1\UiTiDv1ClusterSDK;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+final class BlockConsumers implements ShouldQueue
+{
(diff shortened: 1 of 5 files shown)


app/UiTiDv1/Listeners/BlockConsumers.php
```diff
@@ -0,0 +1,48 @@
...
+    public function __construct(
+        private readonly UiTiDv1ClusterSDK $clusterSDK,
+        private readonly UiTiDv1ConsumerRepository $consumerRepository,
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
+    public function handle(IntegrationBlocked $integrationBlocked): void
+    {
+        $uiTiDv1Consumers = $this->consumerRepository->getByIntegrationId($integrationBlocked->id);
```
bertramakers, line 28:
```suggestion
        $consumers = $this->consumerRepository->getByIntegrationId($integrationBlocked->id);
```
Super small nit, but this would avoid the weird `$uiTiD` spelling (for which there is no good solution I think) and I think the "UiTiD v1" context is implied here (plus we only use the term "consumers" in UiTiD v1 context)

app/UiTiDv1/Listeners/BlockConsumers.php
```diff
@@ -0,0 +1,48 @@
...
+
+        $this->clusterSDK->blockConsumers(...$uiTiDv1Consumers);
+
+        $this->logger->info(
+            'UiTiD v1 consumer(s) blocked',
+            [
+                'domain' => 'insightly',
```
bertramakers, line 35:
```suggestion
                'domain' => 'uitid',
```
I think?

PR #234, opened 2023-03-21
PPF-173 Update UiTiD v1 consumers when integration is updated

LucWollants:
### Changed
- Add listener `UpdateConsumers` which is bound to `IntegrationUpdated` and will update the name and description of the consumer according to the updated integration.

### Note
- The name could be changed depending on the outcome of https://jira.uitdatabank.be/browse/PPF-168

---
Ticket: https://jira.uitdatabank.be/browse/PPF-173

### DIFF
--- a/app/UiTiDv1/Listeners/UpdateConsumers.php
+++ b/app/UiTiDv1/Listeners/UpdateConsumers.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\UiTiDv1\Listeners;
+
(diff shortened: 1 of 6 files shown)


bertramakers (approved):
From a technical standpoint this all looks good. 👌  As with Auth0 I don't really see a reason to sync the description with UiTiD v1, as publiq platform should be the source of truth where we look up integrations and anything that we sync is something that needs to be maintained in the long run. So if it's not technically required, I wouldn't invest time in it. But can be left as is.

PR #242, opened 2023-03-27
PPF-175 Store subscription price in insightly

LucWollants:
### Added
- Added `SubscriptionCategory` enum
- Added `category` column on the `subscriptions` table
- Added call to update subscription on opportunity and project inside Insightly while creating these resources

---
Ticket: https://jira.uitdatabank.be/browse/PPF-175


bertramakers (approved):
As discussed on Slack this looks good from a technical standpoint, but I think that there is an opportunity to simplify the subscriptions model by making the subscriptions enums themselves as it seems from the ticket that they are pretty fixed and don't need to be resources that can be created/edited/deleted in Nova. Having an enum with 4 values for the various subscriptions looks sufficient now. But to be double checked with Corneel first.

LucWollants:
> As discussed on Slack this looks good from a technical standpoint, but I think that there is an opportunity to simplify the subscriptions model by making the subscriptions enums themselves as it seems from the ticket that they are pretty fixed and don't need to be resources that can be created/edited/deleted in Nova. Having an enum with 4 values for the various subscriptions looks sufficient now. But to be double checked with Corneel first.

I think the refactor of the `SubscriptionPlan` should be done inside a separate pull request, so this can already be merged?

PR #251, opened 2023-03-31
PPF-153 Refactor contacts on integration to be optional

LucWollants:
### Changed
- Refactored contacts on integrations to be optional (this will make it consistent with the new URLs which will also be linked resources)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-153

### DIFF
--- a/app/Console/Commands/Migrations/MigrateProjects.php
+++ b/app/Console/Commands/Migrations/MigrateProjects.php
@@ -128,8 +128,7 @@ final class MigrateProjects extends Command
             $migrationProject->name(),
             $migrationProject->description() !== null ? $migrationProject->description() : '',
             Uuid::fromString(SubscriptionsSeeder::BASIC_PLAN),
-            $migrationProject->status(),
-            []
+            $migrationProject->status()
         );
         $this->integrationRepository->save($integration);
 
--- a/app/Domain/Integrations/Controllers/IntegrationController.php
+++ b/app/Domain/Integrations/Controllers/IntegrationController.php
@@ -76,17 +76,14 @@ final class IntegrationController extends Controller
             $this->currentUser->lastName()
         );
 
-        $integration = new Integration(
+        $integration = (new Integration(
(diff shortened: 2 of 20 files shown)


bertramakers:
While the URLs for the Auth0 clients are optional (because not every integrator will work with user logins), I think having a technical and functional contact for an integration is still always required?

LucWollants:
> While the URLs for the Auth0 clients are optional (because not every integrator will work with user logins), I think having a technical and functional contact for an integration is still always required?

Correct, but those validation rules will be moved to the Intertia controller. So the custom-typed classes like `Integration`, `Contact`, and `Organization` are simple DTOs. They are mainly used inside the domain code that is triggered when events happen, for example, the Auth0 listeners.

The Intertia layer for the Integrator/User front-end will need to implement the validation rules.
So that layer has to make sure that:
- An integration has 1 technical and functional contact
- An integration has the correct combination of URLs

bertramakers:
Does that then mean that an admin in Nova could add a second technical or functional contact to an integration, or remove them and leave an integration without technical/functional contact? Because that feels kind of error-prone, especially when the frontend probably won't be able to handle it when that happens 🤔

LucWollants:
> Does that then mean that an admin in Nova could add a second technical or functional contact to an integration, or remove them and leave an integration without technical/functional contact? Because that feels kind of error-prone, especially when the frontend probably won't be able to handle it when that happens 🤔

Correct, there are no validation rules inside Nova for this. It is even not possible to do this in Nova. Because first the integration needs to be created and once the integration is created contacts can be added to it.

LucWollants:
@bertramakers I found a solution to make the technical and functional contact required when creating a new integration. This is done by adding the `Contact` resource with a `hasOne` field which uses an extra accessor on the `IntegrationModel`. The default value of the contact type is set based on the relationship query parameter.

bertramakers (approved):
Nice find 👌

PR #331, opened 2023-05-30
Update frontend dependencies

simon-debruijn:
### Changed

- Update frontend dependencies (see commits)


LucWollants (approved):
Did dependabot not take care of this? Dependabot runs weekly.

simon-debruijn:
> Did dependabot not take care of this? Dependabot runs weekly.

Most of the dependecies where very up to date indeed 😄 Dependabot did a good job 💪🏻 

This is just a manual bump since we're going to work on the code tomorrow.
Also I moved 1 dependency from normal to dev dependency.

PR #347, opened 2023-06-07
PPF-167-navigation

vhande:
### Added
- HeaderMobile
- NavigationElements
- SecondaryButton

Desktop:
HeaderDesktop
Mobile:
HeaderMobile

localhost_(iPhone SE)

---
Ticket: https://jira.uitdatabank.be/browse/...


resources/ts/Shared/NavigationElements.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Link href="/">
+        <Heading
+          className="max-md:text-2xl py-2 border-transparent border-b-4"
+          level={3}
+        >
+          {t("title")}
+        </Heading>
+      </Link>
+      <div className="flex max-md:flex-col md:gap-8 min-w-[50%]">
+        {["integrations", "support"].map((pageTitle) => (
```
brampauwelyn, line 77: Maybe you can define the pages in a `const` above? 
```javascript
const pages = ["integrations", "support"]
```

resources/ts/Shared/SecondaryButton.tsx
```diff
@@ -1,18 +1,24 @@
...
 
-type Props = InertiaLinkProps;
+import { classNames } from "../utils/classNames";
 
-export const SecondaryButton = ({ children, ...props }: Props) => {
+type Props = {
+  orientation?: string;
+} & ComponentProps<"button">;
+
+export const SecondaryButton = ({ children, className, ...props }: Props) => {
```
brampauwelyn, line 9: I would prefer to invert the naming (both in the filename as in the variable name).

```suggestion
export const ButtonSecondary = ({ children, className, ...props }: Props) => {
```

resources/ts/Shared/SecondaryLinkButton.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Link, InertiaLinkProps } from "@inertiajs/react";
+import { classNames } from "../utils/classNames";
+
+type Props = InertiaLinkProps;
+
+export const SecondaryLinkButton = ({
```
brampauwelyn, line 7: Same here I prefer to invert the naming. 
So you only need to read the first part of the file to know that it's about a button...

```suggestion
export const ButtonLinkSecondary = ({
```

simon-debruijn (changes requested):
In mobile when we show the modal/dialog, we should maybe show a darker backdrop behind it. Now the modal doesn't stand out enough compared to the content behind it.

simon-debruijn (approved):

PR #371, opened 2023-06-21
Fix entry form

simon-debruijn:
### Fixed
- Fix entry form


LucWollants (approved):

LucWollants:
@simon-debruijn I already merged so it would be part of the deploy

PR #389, opened 2023-07-04
PPF-187-detail page

vhande:
### Added
- Detail page
- FormDropdown
- BasicInfo
- BillingInfo
- ContactInfo
- IntegrationInfo
- IntegrationSettings

### Changed
-Input
-FormElement
-Layout
-Page

Desktop:
localhost_nl_integraties_a9cb4a29-867c-417e-8239-e3bd1f729321

Mobile:
localhost_nl_integraties_a9cb4a29-867c-417e-8239-e3bd1f729321(iPhone XR)

---


resources/ts/Components/ButtonIcon.tsx
```diff
@@ -15,6 +15,7 @@ export const ButtonIcon = ({
   color,
   size,
   className,
+  onClick,
```
simon-debruijn, line 18: Looks like it should work with `...props` like it was before

resources/ts/Components/FormDropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {
+  faPencil,
+  faChevronRight,
+  faChevronDown,
+} from "@fortawesome/free-solid-svg-icons";
+type Props = {
+  children: React.ReactNode;
+  title: "string";
+  setIsDisabled: React.Dispatch<React.SetStateAction<boolean>>;
+};
```
simon-debruijn, line 13:
```suggestion
} from "@fortawesome/free-solid-svg-icons";

type Props = {
  children: React.ReactNode;
  title: "string";
  setIsDisabled: React.Dispatch<React.SetStateAction<boolean>>;
};
```

resources/ts/Components/FormDropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const FormDropdown = ({ title, children, setIsDisabled }: Props) => {
```
simon-debruijn, line 15: Passing setters like here `setIsDisabled` is a bas practice. It is better to pass 2 props:
- disabled
- onChangeDisabled

simon-debruijn: I'm also wondering.

This disabled state is to collapse and uncollapse right?

Then naming it `isCollapsed` and `setIsCollapsed` seems more correct 🙂

simon-debruijn: But I also see
`const [isVisible, setIsVisible] = useState(false);`

So that could also be your collapsed, which is perfectly fine 👍🏻

resources/ts/Pages/Integrations/Detail.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const [isMobile, setIsMobile] = useState(false);
+
+  const handleResize = () => {
+    setIsMobile(window.innerWidth < 768);
+  };
+
+  useEffect(() => {
+    window.addEventListener("resize", handleResize);
+  });
```
simon-debruijn, line 29: 2 remarks:

1. Make sure to always include a dependency array. Otherwise the code inside will continuously run.
2. return a cleanup function to remove the event listener when the component unmounts

```suggestion
  useEffect(() => {
    window.addEventListener("resize", handleResize);
    return () => window.removeEventListener("resize", handleResize);
  }, []);
```

simon-debruijn (changes requested):

app/Domain/Integrations/Integration.php
```diff
@@ -27,7 +27,7 @@ public function __construct(
         $this->urls = [];
     }
 
-    public function withContacts(Contact ...$contacts): self
+    public function withContacts(Contact...$contacts): self
```
simon-debruijn, line 30:
```suggestion
    public function withContacts(Contact ...$contacts): self
```
I would expect the formatting to look like this

app/Domain/Integrations/Integration.php
```diff
@@ -42,7 +42,7 @@ public function contacts(): array
         return $this->contacts;
     }
 
-    public function withUrls(IntegrationUrl ...$urls): self
+    public function withUrls(IntegrationUrl...$urls): self
```
simon-debruijn, line 45:
```suggestion
    public function withUrls(IntegrationUrl ...$urls): self
```

I would expect the formatting to look like this

simon-debruijn (approved):
The PR is not yet totally finished. It is becoming too large, so we'll merge this one early and make a follow-up PR to make the functionality feature-complete.

PR #453, opened 2023-08-17
PPF-178 Send subscription to insightly when updating project/opportunity

grubolsch:
### Added
- Call updateSubscription when updating a project/opportunity.
- Added tests

### Fixed
- Filter custom fields Insightly instead of just appending 
---
Ticket: https://jira.uitdatabank.be/browse/PPF-178


LucWollants (approved):
Nice work!

PR #474, opened 2023-08-31
PPF-186 Admin can now block individual auth0 clients

grubolsch:
### Added
Admin can now block individual auth0 clients

---


simon-debruijn (approved):

app/Nova/Actions/Auth0/BlockAuth0Client.php
```diff
@@ -0,0 +1,31 @@
...
+    use Queueable;
+
+    public function handle(ActionFields $fields, ActionModelCollection $actionModelCollection): void
+    {
+        foreach ($actionModelCollection as $auth0ClientModel) {
+            if (!$auth0ClientModel instanceof Auth0ClientModel) {
+                continue;
+            }
+
+            ClientBlocked::dispatch(Uuid::fromString($auth0ClientModel->id));
```
LucWollants, line 28: This should become a [Laravel Job](https://laravel.com/docs/10.x/queues#class-structure) instead of an event. The actual block of the client did not happen, this is just the intention from the user.

So the flow is:
- Create a Job called `BlockClient` with a handle method
- Dispatch the `BlockClient` job here
- Inside the handle method of `BlockClient` the actual event `ClientBlocked` can be dispatched

For now we will not listen to `ClientBlocked` but it can be used in the future to update for example the internal state of a client

LucWollants (changes requested):
There is a difference between an `Event`, something that happened and was persisted, and a `Command` or `Job` which is the intention to modify.

app/Auth0/Events/BlockClient.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Auth0\Events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use Ramsey\Uuid\UuidInterface;
+
+final class BlockClient
```
LucWollants, line 10: Maybe move this to the `Jobs` namespace?

app/Auth0/Repositories/EloquentAuth0ClientRepository.php
```diff
@@ -46,6 +46,19 @@ public function getByIntegrationId(UuidInterface $integrationId): array
             ->toArray();
     }
 
+    public function getById(UuidInterface $id): ?Auth0Client
+    {
+        $auth0ClientModel = Auth0ClientModel::query()
```
LucWollants, line 51: I would use a `findOrFail` here. Because calling this code with a wrong `id` is something that needs fixing.

app/Nova/ActionGuards/Auth0/BlockAuth0ClientGuard.php
```diff
@@ -0,0 +1,27 @@
...
+use App\Nova\ActionGuards\ActionGuard;
+
+final class BlockAuth0ClientGuard implements ActionGuard
+{
+    public function __construct(
+        private readonly Auth0ClusterSDK $clusterSDK,
+    ) {
+    }
+
+    public function canDo(object $oAuth0Client): bool
```
LucWollants, line 18: I think it is `auth0Client`

LucWollants (approved):
Nice work! Thanks for the extra effort and time to refactor it to a command/job.

PR #475, opened 2023-08-31
PPF-186 - Admin can now activate individual auth0 clients

grubolsch:
### Added
Admin can now activate individual auth0 clients

---


LucWollants (changes requested):
Same remark as for the block action. There is a difference between an `Event`, something that happened and was persisted, and a `Command` or `Job` which is the intention to modify.

grubolsch:
> Same remark as for the block action. There is a difference between an `Event`, something that happened and was persisted, and a `Command` or `Job` which is the intention to modify.

I have updated the PR with your feedback.

app/Auth0/Jobs/ClientActivatedListener.php
```diff
@@ -0,0 +1,48 @@
...
+            $this->logger->error(
+                'Failed to block Auth0 client',
+                [
+                    'domain' => 'auth0',
+                    'message' => $e->getMessage(),
+                    'id' => $event->id,
+                ]
+            );
+        }
+        $this->logger->info(
```
LucWollants, line 38: I think this success message is traced even when the model was not found.

app/Auth0/Jobs/ClientActivatedListener.php
```diff
@@ -0,0 +1,48 @@
...
+namespace App\Auth0\Jobs;
+
+use App\Auth0\Auth0ClusterSDK;
+use App\Auth0\Events\ClientActivated;
+use App\Auth0\Repositories\Auth0ClientRepository;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Psr\Log\LoggerInterface;
+
+final class ClientActivatedListener implements ShouldQueue
```
LucWollants, line 14: The name should be `ActivateClientListener`

PR #477, opened 2023-08-31
UID-875 Added a cookie that exposes client id when logged in with auth0

grubolsch:
### Added
On the request of @simon-debruijn I exposed the client id of a auth0 user to the frontend with a cookie.

Ticket: https://jira.uitdatabank.be/browse/UID-875

### DIFF
--- a/app/Domain/Auth/Controllers/Callback.php
+++ b/app/Domain/Auth/Controllers/Callback.php
@@ -9,6 +9,7 @@ use Auth0\SDK\Auth0;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Cookie;
 
 final class Callback
 {
@@ -21,7 +22,11 @@ final class Callback
             $user = $auth0->getUser();
 
             if ($user !== null) {
+                Cookie::queue('auth.token.idToken', $auth0->getIdToken(), 3600 * 60 * 24);
+
                 Auth::login(UserModel::fromSession($user));
+            } else {
+                Cookie::queue('auth.token.idToken', '', 0);
             }
         }
 
--- a/app/Domain/Auth/Controllers/Callback.php
+++ b/app/Domain/Auth/Controllers/Callback.php
@@ -22,10 +22,12 @@ final class Callback
             $user = $auth0->getUser();
 
             if ($user !== null) {
-                Cookie::queue('auth.token.idToken', $auth0->getIdToken(), 3600 * 60 * 24);
+                // httpOnly = false
+                Cookie::queue('auth.token.idToken', $auth0->getIdToken(), 3600 * 60 * 24, null, null, null, false);
 
                 Auth::login(UserModel::fromSession($user));
             } else {
+                // Delete the cookie
                 Cookie::queue('auth.token.idToken', '', 0);
             }
         }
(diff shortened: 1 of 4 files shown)


app/Domain/Auth/Controllers/Callback.php
```diff
@@ -21,7 +22,11 @@ public function __invoke(Request $request): RedirectResponse
             $user = $auth0->getUser();
 
             if ($user !== null) {
+                Cookie::queue('auth.token.idToken', $auth0->getIdToken(), 3600 * 60 * 24);
```
simon-debruijn, line 25: We should also disable httpOnly for the frontend to be able to read the cookie

simon-debruijn (changes requested):

app/Domain/Auth/Controllers/Callback.php
```diff
@@ -21,7 +22,13 @@ public function __invoke(Request $request): RedirectResponse
             $user = $auth0->getUser();
 
             if ($user !== null) {
+                // httpOnly = false
+                Cookie::queue('auth.token.idToken', $auth0->getIdToken(), 3600 * 60 * 24, null, null, null, false);
+
                 Auth::login(UserModel::fromSession($user));
+            } else {
+                // Delete the cookie
+                Cookie::queue('auth.token.idToken', '', 0);
```
simon-debruijn, line 31:
```suggestion
                Cookie::expire('auth.token.idToken');
```

I think you can use this method

simon-debruijn (commented):
I think we can write it a bit cleaner like this:

```php
if ($user !== null) {
        $idTokenCookie = Cookie::make(
                name: 'auth.token.idToken',
                value: $auth0->getIdToken() ?? '',
                minutes: 3600 * 60 * 24,
                httpOnly: false
         );
        Cookie::queue($idTokenCookie);

        Auth::login(UserModel::fromSession($user));
} else {
        Cookie::expire('auth.token.idToken');
}
```

app/Domain/Auth/Controllers/Callback.php
```diff
@@ -21,7 +22,13 @@ public function __invoke(Request $request): RedirectResponse
             $user = $auth0->getUser();
 
             if ($user !== null) {
+                // httpOnly = false
+                Cookie::queue('auth.token.idToken', $auth0->getIdToken(), 3600 * 60 * 24, null, null, null, false);
```
LucWollants, line 26: I'm not sure this is the correct location to create the cookie. Because the callback is only reached when the user needs to log in. If we start working with a remember me then this callback endpoint will not be reached.

I think some kind of middleware is a better place.

grubolsch: moved it to the middleware (and learned about middleware in Laraval at the same time)

LucWollants (approved):

app/Http/Middleware/SetAuthIdTokenCookie.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+final class SetAuthIdTokenCookie
+{
+    public const COOKIE = 'auth.token.idToken';
+
+    public function handle(Request $request, Closure $next): Response
+    {
+        /** @var Auth0 $auth0 */
+        $auth0 = app(Auth0::class);
```
simon-debruijn, line 20: I'm guessing there is no better way to inject the Auth0 class here?

simon-debruijn (approved):

PR #487, opened 2023-09-05
PPF-186 Allows activation and blocking of individual uitid clients

grubolsch:
### Added
Allows activation and blocking of individual uitid clients

---


app/UiTiDv1/UiTiDv1ConsumerStatus.php
```diff
@@ -0,0 +1,23 @@
...
+
+enum UiTiDv1ConsumerStatus: string
+{
+    case Active = 'ACTIVE';
+    case Blocked = 'BLOCKED';
+    case Unknown = 'UNKNOWN';
+
+    public static function fromValue(string $name): self
+    {
+        foreach (self::cases() as $status) {
```
LucWollants, line 15: You could use `tryFrom` and when that method returns `null` use `Unknown`.
See https://www.php.net/manual/en/backedenum.tryfrom.php

app/Nova/ActionGuards/UitIdv1/ActivateUitIdv1ClientGuard.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace App\Nova\ActionGuards\UitIdv1;
+
+use App\Nova\ActionGuards\ActionGuard;
+use App\UiTiDv1\UiTiDv1ClusterSDK;
+use App\UiTiDv1\UiTiDv1Consumer;
+use App\UiTiDv1\UiTiDv1ConsumerStatus;
+
+final class ActivateUitIdv1ClientGuard implements ActionGuard
```
LucWollants, line 12: We could start using `readonly classes` from PHP 8.2, see https://stitcher.io/blog/readonly-classes-in-php-82

Maybe a good enhancement to pick up in a follow-up pull request?

app/Nova/ActionGuards/UitIdv1/ActivateUitIdv1ClientGuard.php
```diff
@@ -0,0 +1,28 @@
...
+    ) {
+    }
+
+    public function canDo(object $resource): bool
+    {
+        if (!$resource instanceof UiTiDv1Consumer) {
+            return false;
+        }
+
+        return $this->sdk->fetchStatusOfConsumer($resource) === UiTiDv1ConsumerStatus::Blocked;
```
LucWollants, line 25: Not sure if an activate should be allowed on more status values, maybe a better check is `!==  UiTiDv1ConsumerStatus::Active`

grubolsch: Would rather do not do this. The current other possible value is "unknown", which probably means something went wrong with the api-call, for example the key does not actually exist. Showing an activate button then would be confusing, because it could never work.
I can check for not active and not unknown as alternative?

LucWollants (changes requested):
In the UiTiD v1 context we talk about consumers and not clients. So I would keep that naming convention.

`ActivateClient` => `ActivateConsumer`
`BlockClient` => `BlockConsumer`

app/Nova/Resources/UiTiDv1.php
```diff
@@ -90,4 +95,30 @@ private function getActionUrlTemplates(): array
             )
         );
     }
+
+    public function actions(NovaRequest $request): array
+    {
+        return [
+            (new ActivateUitIdv1Client())
```
LucWollants, line 102: I would also add a confirmation text here (same remark for the Auth0 activate action)

app/Nova/ActionGuards/UitIdv1/ActivateUitIdv1ClientGuard.php
LucWollants: I checked the docs and `ACTIVE` and `BLOCKED` are the only possible states, so keep this check indeed.

That also makes me wonder why we need `UNKNOWN`

LucWollants: If there was something wrong with the API call we should get an exception.

app/Nova/ActionGuards/UitIdv1/ActivateUitIdv1ClientGuard.php
grubolsch: Changed it for my ticket, but there are many classes left to be a candidate for this change.

app/UiTiDv1/UiTiDv1ConsumerStatus.php
grubolsch: cool, did not know that

grubolsch:
> In the UiTiD v1 context we talk about consumers and not clients. So I would keep that naming convention.
> 
> `ActivateClient` => `ActivateConsumer` `BlockClient` => `BlockConsumer`

done

LucWollants (approved):

PR #490, opened 2023-09-07
PPF-197 added logging

grubolsch:
### Added
added logging

---
Ticket: https://jira.uitdatabank.be/browse/PPF-197


app/Auth0/CachedAuth0ClientGrants.php
```diff
@@ -16,6 +18,7 @@ public function findGrantsOnClient(Auth0Client $auth0Client): array
     {
         if(! isset($this->grants[$auth0Client->clientId])) {
             $this->grants[$auth0Client->clientId] = $this->sdk->findGrantsOnClient($auth0Client);
+            Log::debug('CachedAuth0ClientGrants  - ' . $auth0Client->clientId . ': cache miss');
```
LucWollants, line 21: I think a cache hit is more important because a cache miss is already logged through `findGrantsOnClient`

simon-debruijn:
```suggestion
            Log::debug(self::class . '  - ' . $auth0Client->clientId . ': cache miss');
```

You could use the class name intead

LucWollants (approved):

PR #499, opened 2023-09-12
Refactor IntegrationController

simon-debruijn:
### Added

- IntegrationUrlRepository
- Tests for `Domain\Integrations\Mappers`
- Tests for `Domain\Integrations\Repositories\EloquentIntegrationUrlRepository`

### Changed

- better match (naming)conventions of Laravel
- better separation of CRUD operations on integration urls
- Simplify and change formatting for some code

### Removed

- Request specifics in repositories

Another PR will be created to change the frontend code to use the new endpoints.


app/Domain/Integrations/Mappers/UpdateContactInfoMapper.php
```diff
@@ -68,9 +68,9 @@ public static function map(UpdateContactInfo $updateContactInfo, string $integra
             $contacts[] = $contact;
         }
 
-        $newLastName = $updateContactInfo->input('newContributorLastName');
-        $newFirstName = $updateContactInfo->input('newContributorFirstName');
-        $newEmail = $updateContactInfo->input('newContributorEmail');
+        $newLastName = $request->input('newContributorLastName');
+        $newFirstName = $request->input('newContributorFirstName');
+        $newEmail = $request->input('newContributorEmail');
```
simon-debruijn, line 73: I'll also move this logic to a separate `POST` call to `/integrations/{id}/contacts`

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -77,20 +86,36 @@ public function create(): Response
...
-            TranslatedRoute::getTranslatedRouteName(
-                request: $storeIntegration,
-                routeName: 'integrations.index'
-            )
+            TranslatedRoute::getTranslatedRouteName($request, 'integrations.index')
+        );
+    }
+    public function storeUrl(StoreIntegrationUrlRequest $request, string $id): RedirectResponse
+    {
+        $integrationUrl = new IntegrationUrl(
```
LucWollants, line 100: Although it's a "oneline" I think a mapper called `StoreIntegrationUrlMapper` could be nice here.

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -151,45 +212,39 @@ public function deleteContact(Request $request, string $id, string $contactId):
...
 
-    public function updateBilling(string $id, UpdateBillingInfo $updateBillingInfo): RedirectResponse
+    public function updateBilling(string $id, UpdateBillingInfoRequest $request): RedirectResponse
     {
         $organisation = new Organization(
-            Uuid::fromString($updateBillingInfo->input('organisation.id')),
-            $updateBillingInfo->input('organisation.name'),
+            Uuid::fromString($request->input('organisation.id')),
+            $request->input('organisation.name'),
             'test@test.be',
```
LucWollants, line 227: Is this `test@test.be` intentional or a typo?

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -151,45 +212,39 @@ public function deleteContact(Request $request, string $id, string $contactId):
...
             [
                 'id' => $id,
             ]
         );
     }
 
-    public function updateBilling(string $id, UpdateBillingInfo $updateBillingInfo): RedirectResponse
+    public function updateBilling(string $id, UpdateBillingInfoRequest $request): RedirectResponse
     {
         $organisation = new Organization(
```
LucWollants, line 224: Another candidate for a mapper?

app/Domain/Integrations/FormRequests/StoreIntegrationUrlRequest.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Foundation\Http\FormRequest;
+
+final class StoreIntegrationUrlRequest extends FormRequest
+{
+    /**
+     * @return array<string, mixed>
+     */
+    public function rules(): array
+    {
+        return [
```
LucWollants, line 16: Not sure how stricted these rules should be, but should we also verify if:
- The environment is known?
- The type is known?
- The url is a valid HTTPS url?

app/Domain/Integrations/FormRequests/UpdateIntegrationUrlsRequest.php
```diff
@@ -0,0 +1,29 @@
...
+    public function rules(): array
+    {
+        $urlValidation = [
+            'id' => ['required', 'string'],
+            'url' => ['required_without:environment', 'string', 'max:255'],
+            'environment' => ['required_without:url', 'string', 'max:255'],
+        ];
+
+        return [
+            'loginUrls.*' => Rule::forEach(fn () => $urlValidation),
```
LucWollants, line 24: Only one login URL is allowed

app/Domain/Integrations/Mappers/UpdateContactInfoMapper.php
LucWollants: We already have `IntegrationController::updateContacts`, so I think that is fine?

app/Domain/Integrations/Repositories/EloquentIntegrationUrlRepository.php
```diff
@@ -0,0 +1,90 @@
...
+
+    public function updateUrls(array $integrationUrls): void
+    {
+        if (count($integrationUrls) === 0) {
+            return;
+        }
+
+        DB::transaction(static function () use ($integrationUrls) {
+            foreach ($integrationUrls as $integrationUrl) {
+                IntegrationUrlModel::query()->update(
```
LucWollants, line 65: You could directly call the existing `EloquentIntegrationUrlRepository::update` method

app/Domain/Integrations/Repositories/IntegrationUrlRepository.php
```diff
@@ -0,0 +1,25 @@
...
+    /**
+    * @param array<IntegrationUrl> $integrationUrls
+     */
+    public function updateUrls(array $integrationUrls): void;
+    public function getById(UuidInterface $id): IntegrationUrl;
+    /**
+    * @param array<UuidInterface> $ids
+    * @return array<IntegrationUrl>
+     */
+    public function getByIds(array $ids): array;
```
LucWollants, line 23: I noticed we have a mix of `PaginatedCollection` and `array`. To be consistent I would replace the `array` with `Collection`.

Maybe something for a follow-up pull request?

app/Domain/Integrations/Mappers/UpdateContactInfoMapper.php
simon-debruijn: Yes, it works. We can send `newContributorLastName`, ... to the `updateContacts`.
Wouldn't it be cleaner if the creation of a new contact was separate?

app/Domain/Integrations/Controllers/IntegrationController.php
simon-debruijn: A remaining dummy value, will fix it

app/Domain/Integrations/FormRequests/StoreIntegrationUrlRequest.php
simon-debruijn: Yes 👍🏻 For storing and updating aswell

app/Domain/Integrations/Mappers/UpdateContactInfoMapper.php
LucWollants: I think I now understand de confusion. This is a mix between new contacts and updated contacts. I think it was designed that way to better match the UI that has a big editable list.

If I would create an endpoint without taking those requirements into account I would indeed make a store and update endpoint. The consumer of those endpoints needs to call the store and update one or several times.

app/Domain/Integrations/FormRequests/StoreIntegrationUrlRequest.php
simon-debruijn: `environment` and `type` will fail on the conversion to its enum value:

```php
Environment::from($request->input('environment')),
IntegrationUrlType::from($request->input('type')),
```

The frontend also only let's you choose the supported values.

So in a worst case scenario if someone tries to send their own modified inertia request, it will fail right after the laravel validation at the enum conversion.

app/Domain/Integrations/Repositories/IntegrationUrlRepository.php
simon-debruijn: Let's do it in a follow-up PR

app/Domain/Integrations/Mappers/UpdateContactInfoMapper.php
simon-debruijn: In the frontend we are changing the ui atm to use dialogs/modals for creating a new contact for example. There it makes sense to fire a POST request solely for that new contact.

LucWollants: Yes, I would also prefer a clear `POST` endpoint to create new contacts and a `PUT` endpoint to update existing contacts.

LucWollants: But to limit the scope of this PR it may be best to do this in a follow-up pull request?

simon-debruijn:
I've added tests for the mapper classes and the integration url repository

grubolsch (approved):

grubolsch:
Approved, but like discussed make a new ticket / PR to chance a lot of the unitTests to use TestCase instead of TestCaseWithDatabase, because they don't need the DB connection.
This will make the test suite faster.

PR #519, opened 2023-09-25
PPF-187-update detail page

vhande:
### Added
- ContributorTable component
- Tabs component
- Delete contributor function

### Changed
- Dialog style

---


simon-debruijn (approved):

simon-debruijn (changes requested):
I still need to review, approval was accidental 😉

resources/ts/Components/ContributorTable.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { ComponentProps } from "react";
+import { useTranslation } from "react-i18next";
+
+type Props = ComponentProps<"div">;
+
+export default function ContributorTable({ children }: Props) {
```
simon-debruijn, line 6:
```suggestion
export const ContributorTable = ({ children }: Props) => {
```

resources/ts/Components/Dialog.tsx
```diff
@@ -36,7 +37,12 @@ export const Dialog = ({
           className="text-publiq-blue-dark self-end"
         />
 
-        <div className="flex flex-col flex-1 w-full p-4 text-xl">
+        <div
+          className={classNames(
+            "flex flex-col flex-1 w-full p-4 text-xl",
+            className
```
simon-debruijn, line 43: It's misleading that className is not used in the root of this component.
You could introduce a prop `contentStyles` or something similar and pass it here

resources/ts/Pages/Integrations/Index.tsx
```diff
@@ -95,7 +95,7 @@ const Index = ({ integrations, paginationInfo }: Props) => {
           search: e.target.value || undefined,
         },
         {
-          preserveState: true,
+          // preserveState: true,
```
simon-debruijn, line 98: This can be removed

simon-debruijn (changes requested):

PR #524, opened 2023-09-29
PPF-201-navigation-link

vhande:
### Added
- Link to integration entry form in menu bar

---
Ticket: https://jira.uitdatabank.be/browse/PPF-201


resources/ts/Components/Navigation.tsx
```diff
@@ -37,9 +37,7 @@ export default function Navigation({
   const { t, i18n } = useTranslation();
   const translateRoute = useTranslateRoute();
 
-  const path = new URL(document.location.href).pathname;
-
-  const pages = ["integrations", "support"];
+  const pages = ["integrations", "integrations/new", "support"];
 
   const { component } = usePage();
```
simon-debruijn, line 42: Can't we just usse this to infer the current page and compare it to the links instead of keeping active link in state?

simon-debruijn (changes requested):

resources/ts/Components/Navigation.tsx
vhande: Since we use the startsWith() method and both strings start with "integrations", it doesn't seem to be working.

simon-debruijn (approved):

PR #525, opened 2023-09-29
Add default IntegrationPartnerStatus to migration up

simon-debruijn:
### Added

- Add default IntegrationPartnerStatus to migration up

### Fixed

- displaying of integrations failing because `""` is not a valid IntegrationPartnerStatus


LucWollants:
Because this migration was already deployed, a force migration is needed on the ACC server.

vhande (approved):

PR #533, opened 2023-10-06
PPF-187-detail page

vhande:
### Added
-ContactsTable
-ContactsTableContent
-ContactsTableDesktop
-ContactsTableMobile

### Changed
-Card
-IntegrationCard
-IntegrationSettings
-Dialog
-Translations

### Removed
-ContributorTable

contactstable
integrationsettings
integrationcard
modal1
modal2

---


resources/ts/Components/Card.tsx
```diff
@@ -8,28 +8,33 @@ type Props = {
...
 export const Card = ({
   title,
   description,
   img,
+  badge,
   active,
   children,
   className,
   contentStyles,
+  border,
```
simon-debruijn, line 27: The props that are optional and boolean like `border` and `badge` better have a explicit false default value

resources/ts/Components/ContactsTableContent.tsx
```diff
@@ -0,0 +1,146 @@
+import React, { ComponentProps } from "react";
+import { ContactsTableProps } from "./ContactsTable";
+import { classNames } from "../utils/classNames";
+import { ButtonIcon } from "./ButtonIcon";
+import { faPencil, faTrash } from "@fortawesome/free-solid-svg-icons";
+import { useTranslation } from "react-i18next";
+import { ContactFormData } from "./Integrations/Detail/ContactInfo";
+import { Contact } from "../Pages/Integrations/Index";
+type Props = { desktop?: boolean; mobile?: boolean } & ContactsTableProps &
+  ComponentProps<"div">;
```
simon-debruijn, line 10:
```suggestion
import { Contact } from "../Pages/Integrations/Index";

type Props = { desktop?: boolean; mobile?: boolean } & ContactsTableProps &
  ComponentProps<"div">;
```

resources/ts/Components/ContactsTableContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export const ContactsTableContent = ({
+  desktop,
+  mobile,
```
simon-debruijn, line 14: You could also use a display prop instead with type `'mobile' | 'desktop'`
But this approach is also perfect 👍🏻

resources/ts/Components/ContactsTableContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+          >
+            <div
+              className={classNames(
+                desktop && "px-6 pt-4 font-semibold",
+                mobile &&
+                  "text-publiq-blue-dark flex gap-2 items-center underline"
+              )}
+            >
+              {
+                (data[type.label as keyof ContactFormData] as Contact)
```
simon-debruijn, line 60: We can probably search the correct contact in the mapping of contactTypes

Something like this:

```ts
{contactTypes.map((type) => {
    const contact = data[type.label];
    
    return (
        <tr
```

Also you can change
```ts
const contactTypes = [
    { label: "functional", id: functionalId },
    { label: "technical", id: technicalId },
  ];
```

To

```ts
const contactTypes = [
    { label: "functional", id: functionalId },
    { label: "technical", id: technicalId },
  ] as const;
```

You won't need the `as keyof ContactFormData`

simon-debruijn (changes requested):

brampauwelyn:
@vhande 
One little design remark.
I would make the border of the whole card less hard. 
Maybe use a lighter grey? Or no border at all with a slight box shadow?

simon-debruijn (approved):

PR #545, opened 2023-10-11
PPF-212 - optimize tests

simon-debruijn:
### Changed
- Make TestCase as simple as possible
- Use `Tests\TestCase` everywhere
- Call RefreshDatabase in specific tests

---
Ticket: https://jira.uitdatabank.be/browse/PPF-212

The specific benchmarking and testing I did you can find in the comments of the ticket


tests/Auth0/Listeners/BlockClientsTest.php
```diff
@@ -12,7 +12,7 @@
 use App\Json;
 use GuzzleHttp\Psr7\Response;
 use PHPUnit\Framework\MockObject\MockObject;
-use PHPUnit\Framework\TestCase;
```
LucWollants, line 15: Maybe the PHPUnit TestCases can stay as is?

LucWollants (approved):

tests/Auth0/Listeners/BlockClientsTest.php
simon-debruijn: In my testing changing it to the custom testcase didn't slow things down. Isn't it less confusing if we only use 1 testcase everywhere?

LucWollants: Ok, fine for me!

PR #548, opened 2023-10-12
PPF-202-Add clickable heading

vhande:
### Added
-clickable heading

---
Ticket: https://jira.uitdatabank.be/browse/PPF-202


resources/ts/Components/Card.tsx
```diff
@@ -28,8 +32,11 @@ export const Card = ({
   contentStyles,
   border = false,
   textCenter,
+  clickableHeading,
```
simon-debruijn, line 35:
```suggestion
  clickableHeading = false,
```

simon-debruijn (approved):

PR #566, opened 2023-10-19
Added analytics to pages

Anahkiasen:
### Added
- Added analytics to pages

---

I wasn't able to confirm this works as is, since none of my events showed in Realtime or DebugView. However using Google's own tag debugger it said everything was working properly and I did see my events coming in there. The documentation say there's a 24 to 48h period before tracking is correct, even in Realtime, so I'm guessing it's just this?

Ticket: https://jira.publiq.be/browse/PPF-217


.env.example
```diff
@@ -128,3 +128,5 @@ INSIGHTLY_API_KEY=
 
 SENTRY_LARAVEL_ENABLED=true
 SENTRY_LARAVEL_DSN=
+
+GTAG_MEASUREMENT_ID=
```
LucWollants, line 132: I think we need to add this to appconfig repo also https://github.com/cultuurnet/appconfig/tree/main/files/platform

LucWollants (approved):

simon-debruijn (approved):

PR #568, opened 2023-10-20
PPF-225-home product cards

vhande:
### Added
-Custom drop shadow
### Changed
- IntegrationTypeCard style

Cards

---
Ticket: https://jira.uitdatabank.be/browse/PPF-225


simon-debruijn (changes requested):
The drop shadow looks a bit to different from the normal cards. Can you match them, so that they are the same apart from the triangle?

Screenshot 2023-10-24 at 09 48 06
Screenshot 2023-10-24 at 09 48 03

simon-debruijn (approved):

PR #577, opened 2023-10-26
PPF-188-support slack

vhande:
### Added
- SupportController
- InformationDialog

---
Ticket: https://jira.uitdatabank.be/browse/PPF-188


app/Http/Controllers/SupportController.php
```diff
@@ -0,0 +1,57 @@
...
+            'email' => Auth::user()?->email,
+            'slackSuccess' => $slackSuccess,
+            'slackError' => $slackError,
+        ]);
+    }
+    public function sendInvitation(Request $request): RedirectResponse
+    {
+        $botToken = config('slack.botToken');
+        $channelID = config('slack.channelID');
+        $email = Auth::user()?->email;
```
simon-debruijn, line 33: The `sendInvitation` method should probably be under the authenticated routes. If the user is not authenticated and on the support page and clicks the join slack button we first let them login.

Then we'll be sure that `Auth::user()` is not null

resources/translations/nl.json
```diff
@@ -259,7 +259,11 @@
   },
   "dialog": {
     "confirm": "Bevestigen",
-    "cancel": "Annuleren"
+    "cancel": "Annuleren",
+    "close": "Sluiten",
+    "invite_success": "Wij hebben een uitnodiging gestuurd naar {{email}}",
+    "invite_wrong": "Er is iets fout gegaan. Probeer het later opnieuw."
```
simon-debruijn, line 265:
```suggestion
    "invite_error": "Er is iets fout gegaan. Probeer het later opnieuw."
```
For consistency

routes/web.php
```diff
@@ -38,9 +38,12 @@
         '/en/support',
         '/nl/ondersteuning',
     ],
-    static fn () => Inertia::render('Support/Index')
+    [SupportController::class, 'index'],
+    'support.index'
 );
 
+Route::post('/support/slack', [SupportController::class, 'sendInvitation']);
```
simon-debruijn, line 45:
```suggestion
Route::post('/support/slack/invite', [SupportController::class, 'sendSlackInvitation']);
```

simon-debruijn (changes requested):

resources/ts/Components/SupportCard.tsx
```diff
@@ -4,16 +4,31 @@ import { Heading } from "./Heading";
...
   actionTitle,
   actionUrl,
+  email,
+  slackError,
+  slackSuccess,
 }: Props) => {
+  const { t } = useTranslation();
+  const translateRoute = useTranslateRoute();
+  const handleSlackInvitation = () => router.post("/support/slack");
+  const handleRedirect = () => router.get(translateRoute("/support"));
```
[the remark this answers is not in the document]
vhande: It doesn't work as we have one more redirect after this.

app/Http/Controllers/SupportController.php
```diff
@@ -39,7 +39,13 @@ public function sendInvitation(Request $request): RedirectResponse
...
                 ]);
 
-            if (!$response->ok()) {
+            $body = json_decode(
+                json: $response->body(),
+                associative: true,
+                flags: JSON_THROW_ON_ERROR
+            );
+
+            if (!$response->ok() | $body['ok'] === false) {
```
simon-debruijn, line 48:
```suggestion
            if (!$response->ok() || $body['ok'] === false) {
```

simon-debruijn (changes requested):

app/Http/Controllers/SupportController.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $botToken = config('slack.botToken');
+        $channelID = config('slack.channelID');
+        $email = Auth::user()?->email;
+
+        try {
+            $response = Http::withToken($botToken)
+                ->post('https://slack.com/api/conversations.inviteShared', [
+                    'channel' => $channelID,
+                    'emails' => $email,
```
simon-debruijn, line 39: I now see in the slack docs that emails is supposed to be an array. So even if this works, it is better to follow the correct guidelines.

```suggestion
                    'emails' => [$email],
```

app/Http/Controllers/SupportController.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    public function index(Request $request): Response
+    {
+        $slackSuccess = $request->query->has('slackSuccess');
+        $slackError = $request->query->has('slackError');
+
+        return Inertia::render('Support/Index', [
+            'email' => Auth::user()?->email,
+            'slackSuccess' => $slackSuccess,
+            'slackError' => $slackError,
```
simon-debruijn, line 26: We can also merge these into 1 `slackStatus` which can be `success` or `error`

simon-debruijn (approved):

PR #601, opened 2023-11-07
PPF-228-Add GTM for cookie-script

vhande:
### Added
- GTM for cookie-script

---
Ticket: https://jira.uitdatabank.be/browse/PPF-228


resources/views/layouts/default.blade.php
```diff
@@ -12,8 +12,10 @@
         <title>publiq-platform</title>
         @inertiaHead
         @include('layouts.hotjar')
+        @include('layouts.gtm', ['part' => 'head'])
```
brampauwelyn, line 15: I think you can put it before hotjar even

brampauwelyn (approved):

PR #604, opened 2023-11-09
PPF-229-settings tab

vhande:
### Added
-Alert component

### Fixed
-Settings Tab

success
error

---
Ticket: https://jira.uitdatabank.be/browse/PPF-229


app/Domain/Integrations/Models/IntegrationUrlModel.php
```diff
@@ -29,10 +30,17 @@ final class IntegrationUrlModel extends UuidModel
     protected static function booted(): void
     {
         self::created(
-            static fn (IntegrationUrlModel $integrationUrlModel) => IntegrationUrlCreated::dispatch(
-                Uuid::fromString($integrationUrlModel->id),
-                Uuid::fromString($integrationUrlModel->integration_id)
-            )
+            static function (IntegrationUrlModel $integrationUrlModel) {
+                // @phpstan-ignore-next-line
+                $id = $integrationUrlModel->id instanceof LazyUuidFromString ? $integrationUrlModel->id : Uuid::fromString($integrationUrlModel->id);
```
LucWollants, line 35: What is causing this `id` to have type `LazyUuidFromString`?

app/Domain/Integrations/FormRequests/UpdateIntegrationUrlsRequest.php
```diff
@@ -20,9 +21,33 @@ public function rules(): array
         ];
 
         return [
-            'loginUrl' => $urlValidation,
+            'integrationName' => ['required', 'string', 'max:255'],
+            'description' => ['required', 'string', 'max:255'],
+            'loginUrls.*' => Rule::forEach(fn () => $urlValidation),
             'callbackUrls.*' => Rule::forEach(fn () => $urlValidation),
             'logoutUrls.*' => Rule::forEach(fn () => $urlValidation),
+            'newIntegrationUrls.*' => Rule::forEach(fn () => [
```
LucWollants, line 29: What kind of URLs are `newIntegrationUrls`?

app/Domain/Integrations/Models/IntegrationUrlModel.php
simon-debruijn: We investigated this, but it looks like an implementation from the uuid lib detail rather than our code

app/Domain/Integrations/FormRequests/UpdateIntegrationUrlsRequest.php
simon-debruijn: Those are urls that have been added in the frontend in the current usage of the website. 
The state for already existing urls that are being edited is in separate fields from the newly added urls.

app/Domain/Integrations/Models/IntegrationUrlModel.php
LucWollants: Could it be that the new integration URLs sometimes have an empty ID?

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -140,41 +133,61 @@ public function destroyUrl(Request $request, string $id, string $urlId): Redirec
...
-            [
-                'id' => $id,
-            ]
-        );
+            $this->integrationRepository->update($updatedIntegration);
+        }
     }
 
-    public function updateUrls(UpdateIntegrationUrlsRequest $request, string $id): RedirectResponse
+    public function update(UpdateIntegrationUrlsRequest $request, string $id): RedirectResponse
```
LucWollants, line 147: Was this name change intentional? Because the request has still type `UpdateIntegrationUrlsRequest` but the method name makes me think the full integration gets updated?

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -140,41 +133,61 @@ public function destroyUrl(Request $request, string $id, string $urlId): Redirec
...
+        $integrationUpdateRequest = new UpdateIntegrationRequest($integrationRequest);
+
+        $this->updateIntegration($integrationUpdateRequest, $id);
+
         $ids = array_map(
             fn ($url) => Uuid::fromString($url['id']),
             array_filter(
                 [
-                    $request->input('loginUrl'),
+                    ...($request->input('loginUrls') ?? []),
```
LucWollants, line 159: I think in Auth0 there is only a single login URL

simon-debruijn: Yes indeed loginUrl should be singular

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -142,41 +134,61 @@ public function destroyUrl(Request $request, string $id, string $urlId): Redirec
...
 
-        $this->integrationUrlRepository->updateUrls($updatedUrls);
+            $this->integrationUrlRepository->updateUrls($updatedUrls);
+        }
+
+        $newUrls =
+            array_filter(
+                [
+                    ...($request->input('newIntegrationUrls') ?? []),
+                ],
```
simon-debruijn, line 180:
```suggestion
                ($request->input('newIntegrationUrls') ?? []),
```

No need to create a new array

app/Domain/Integrations/FormRequests/UpdateIntegrationUrlsRequest.php
simon-debruijn: We better look into using upsert (update or insert depending on if it exists) logic

app/Domain/Integrations/Mappers/UpdateIntegrationUrlsMapper.php
```diff
@@ -21,7 +21,7 @@ public static function map(UpdateIntegrationUrlsRequest $request, array $current
 
         $changedUrlsCollection = Collection::make(
             [
-                $request->input('loginUrl'),
+                ...($request->input('loginUrls') ?? []),
```
simon-debruijn, line 24: Same remark about loginUrl being singular

resources/ts/Components/FormElement.tsx
```diff
@@ -80,14 +81,18 @@ export const FormElement = ({
   component,
   error,
   info,
+  elementId,
```
simon-debruijn, line 84: Why is this needed?

resources/ts/Components/Integrations/Detail/IntegrationSettings.tsx
```diff
@@ -90,6 +110,14 @@ export const IntegrationSettings = ({ integration, id, urls }: Props) => {
     setData("newIntegrationUrls", updated);
   };
 
+  const handleDeleteNewUrl = (fields?: string[], id?: string) => {
+    const updatedUrls = data.newIntegrationUrls
+      .filter((url) => fields?.includes(url.id))
+      .filter((url) => url.id !== id);
```
simon-debruijn, line 116:
```suggestion
      .filter((url) => fields?.includes(url.id) && url.id !== id);
```

simon-debruijn (changes requested):
I'll take a look how we can further improve simplify it when it is a upw week

resources/ts/Components/Integrations/Detail/UrlList.tsx
```diff
@@ -86,71 +120,108 @@ export const UrlList = ({
...
-                </div>
-              )}
+                    )
+                )}
+              </div>
               {type !== IntegrationUrlType.Login && (
                 <ButtonSecondary
                   onClick={() => {
-                    option.changeVisibility(true);
+                    option.addNewUrlField(`${option.env + randomNumber}`);
```
brampauwelyn, line 224: What is the reason of the `randomNumber`?

resources/ts/Components/Integrations/Detail/UrlList.tsx
```diff
@@ -159,40 +230,107 @@ export const UrlList = ({
...
-                        type,
-                      })
-                    }
-                  />
-                </div>
-              }
-            />
+            <div key={option.env} className="flex flex-col gap-2">
+              <FormElement
+                label={`${t(`details.integration_settings.${option.env}`)}`}
```
brampauwelyn, line 235: You can simplify this a bit:
```suggestion
                label={t(`details.integration_settings.${option.env}`)}
```

resources/ts/Components/Integrations/Detail/UrlList.tsx
```diff
@@ -86,71 +120,108 @@ export const UrlList = ({
...
-                    `details.integration_settings.${url.environment}`
-                  )}`}
+                  elementId={url.id}
+                  className={`${type + option.env}`}
+                  error={errors[`${type}Urls.${`${url.id}`}.url`]}
+                  label={
+                    index === 0
+                      ? `${t(
+                          `details.integration_settings.${url.environment}`
+                        )}`
```
brampauwelyn, line 133:
```suggestion
                      ? t(
                          `details.integration_settings.${url.environment}`
                        )
```

resources/ts/Components/Alert.tsx
```diff
@@ -0,0 +1,88 @@
...
+        "flex gap-3 justify-between items-start rounded-md p-4",
+        variantToAlertStyle[variant],
+        className
+      )}
+    >
+      <div className={"flex gap-3 items-center"}>
+        <FontAwesomeIcon
+          icon={variantToIcon[variant]}
+          className={variantToIconColor[variant]}
+          size={"xl"}
```
brampauwelyn, line 71:
```suggestion
          size="xl"
```

resources/ts/Components/Integrations/Detail/UrlList.tsx
simon-debruijn: The PR isn't ready for review yet, my bad.

This is remaining code I still have to refactor

resources/ts/Components/Integrations/Detail/UrlList.tsx
simon-debruijn: I no longer find this code

resources/ts/Components/Integrations/Detail/UrlList.tsx
simon-debruijn: I no longer find this code

simon-debruijn (approved):

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
```diff
@@ -45,14 +45,14 @@ public function save(Integration $integration): void
...
     {
-        IntegrationModel::query()
-            ->findOrFail($integration->id->toString())
-            ->update([
-                'type' => $integration->type,
-                'name' => $integration->name,
-                'description' => $integration->description,
-                'subscription_id' => $integration->subscriptionId->toString(),
-                'status' => $integration->status,
+        $integrationModel = IntegrationModel::query()->findOrFail($integration->id->toString());
```
LucWollants, line 48: Is there a special need for the local variable `$integrationModel`?

app/Domain/Integrations/FormRequests/UpdateIntegrationUrlsRequest.php
```diff
@@ -14,15 +18,36 @@ final class UpdateIntegrationUrlsRequest extends FormRequest
...
+            'urls.*' => Rule::forEach(fn (array $value, string $attribute) => [
+                'id' => ['string'],
+                'environment' => ["required_without:$attribute.id", new Enum(Environment::class)],
+                'type' => ["required_without:$attribute.id", new Enum(IntegrationUrlType::class)],
+                'url' => ['required', 'url:https', 'max:255'],
+            ]),
         ];
+    }
 
+    public function after(): array
```
LucWollants, line 31: This looks rather complex, by just looking at the code I don't know what it does 😅

simon-debruijn: The after method is called after the main validation is done.

The error keys that will be generated are in the following structure:
`urls.0.url`

We can't map these keys to the frontend fields. Because in the frontend we group and filter the urls based on the type and environment.

What this code does, is create new error keys with the following structure:
`type.environment.url`
We also have to use `getData` to use the correct type, environment and url per url object

And after adding out custom error keys, we forget (delete) the old error keys

simon-debruijn: In the laravel docs it says the after method should be replaced, by something else:
https://laravel.com/docs/10.x/upgrade#form-request-after-method

https://laravel.com/docs/10.x/validation#performing-additional-validation

But to me its unclear what the alternative is.

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
simon-debruijn: I don't think so

simon-debruijn: I see other instances in the code where it also defined first, to then call a method on it.
e.g.
```php
$integrationModel = IntegrationModel::query()->findOrFail($id->toString());
$integrationModel->activateWithOrganization($organizationId);
```

LucWollants: But that line is to convert it a `IntegrationModel` and make PHPStan happy I think.

app/Domain/Integrations/FormRequests/UpdateIntegrationUrlsRequest.php
LucWollants: Thanks for the extra comments. I think the first bigger comment was enough, the rest can be deduced from reading the code.

LucWollants (approved):

PR #635, opened 2023-11-28
Add an endpoint to check if a user has access to an integration

LucWollants:
### Added
- Added an endpoint to check if a user has access to an integration

---
Ticket: https://jira.uitdatabank.be/browse/PPF-236

### DIFF
--- a/app/Domain/Auth/Controllers/Access.php
+++ b/app/Domain/Auth/Controllers/Access.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Domain\Auth\Controllers;
+
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Http\Controllers\Controller;
+use Auth0\SDK\Auth0;
+use Psr\Log\LoggerInterface;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+final class Access extends Controller
+{
+    public function __construct(
+        private readonly IntegrationRepository $integrationRepository,
+        private readonly LoggerInterface $logger
+    ) {
+    }
+
+    public function handle(string $idToken, string $integrationId): JsonResponse
+    {
+        $this->logger->info('Requested IntegrationAccess for integration ' . $integrationId);
+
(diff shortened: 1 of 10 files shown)


JonasVHG (commented):
I keep getting errors while trying to add a contact to an integration. Probably something is missing from the config.

routes/api.php
```diff
@@ -20,3 +21,4 @@
 });
 
 Route::get('token/{idToken}', [Token::class, 'handle']);
+Route::get('token/{idToken}/integration/{id}', [Access::class, 'handle']);
```
simon-debruijn, line 24: Shouldn't we postfix these controller names with Controller like we did in `web.php`?

e.g. `AccessController` and `TokenController`

app/Domain/Auth/Controllers/Access.php
```diff
@@ -0,0 +1,51 @@
...
+        try {
+            $token = $auth0->decode($idToken);
+        } catch (\Exception $exception) {
+            return new JsonResponse(
+                ['exception' => $exception->getMessage()],
+                400
+            );
+        }
+
+        $email = $token->toArray()['email'];
```
simon-debruijn, line 37: We could have a class (UserInfo or TokenPayload) that contains the properties instead of a plain array
It can fail appropriately if an expected property is missing.

Something like
```php
$userInfo = UserInfo::from($token);
$email = $userInfo->email
```

simon-debruijn (approved):

routes/api.php
LucWollants: Those names are indeed not consistent. I will add the `Controller` postfix and also change the already existing ones like `Login, `Logout` and `Callback`

The names `Login`, `Logout`, and `Callback` did follow the Auth0 SDK conventions.

LucWollants:
> I keep getting errors while trying to add a contact to an integration. Probably something is missing from the config.

As discussed through Slack, this extra config is needed https://github.com/cultuurnet/appconfig/pull/504

app/Domain/Auth/Controllers/Access.php
LucWollants: It's only used in the location so I will add some extra logging and a better response.

PR #659, opened 2023-12-11
PPF-227-Merge checkboxes

vhande:
### Changed
- [Use trans component in lieu of clickable label and link](https://github.com/cultuurnet/publiq-platform/commit/d3d8e1dfd08c331634dc9f7c94ca68266c54d723)
- [Merge terms of use and privacy](https://github.com/cultuurnet/publiq-platform/commit/286a438eb965674026528113183c021c0cd0eb74)

### Removed
- [Delete privacy validation](https://github.com/cultuurnet/publiq-platform/commit/e65e317709eb3befabbbad19348bcf7919dac3d9)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-227


resources/ts/Components/FormElement.tsx
```diff
@@ -65,9 +65,7 @@ const InputStyle = {
 };
 
 type Props = {
-  label?: string;
-  clickableLabel?: string;
-  clickableLabelLink?: string;
+  label?: string | ReactElement;
```
LucWollants, line 68: Having two totally different types is mostly a code smell. Maybe this combined field is no longer a label but the component itself?

simon-debruijn: I think `ReactElement` even contains the string type. But would have to check.
For a label I can see the use case where you most of the time want a simple string that will be injected into a label element.
But sometimes we need more control over the element injected into the label. e.g. setting html content with custom link components as it is used here.

simon-debruijn: I looked it up

`ReactElement` doesn't contain string
`ReactNode` does (and all other options react could render)

So hande's type seems the most restrictive option

LucWollants (approved):

PR #692, opened 2024-01-04
Styling improvements

Anahkiasen:
### Changed
- Styling improvements

---

Current status:

Screenshot 2024-01-04 at 13 41 08
Screenshot 2024-01-04 at 13 41 28

Ticket: https://jira.publiq.be/browse/PPF-221


resources/ts/Components/Card.tsx
```diff
@@ -39,13 +39,16 @@ export const Card = ({
...
       className={classNames(
-        "w-full flex flex-col overflow-hidden shadow-lg",
+        "w-full flex flex-col overflow-hidden drop-shadow-card",
         img && "px-0 py-0 gap-10 max-lg:gap-3 p-0",
-        active ? "bg-status-green-medium bg-opacity-10" : "bg-white",
-        border ? "border border-gray-300" : "px-6 py-6",
+        active
+          ? "bg-status-green-medium bg-opacity-10"
+          : "bg-publiq-gray-light",
+        border ? "" : "px-6 py-6",
```
brampauwelyn, line 51:
```suggestion
        border && "px-6 py-6",
```

resources/ts/Components/ButtonIcon.tsx
```diff
@@ -19,9 +20,11 @@ export const ButtonIcon = ({
 }: Props) => {
   return (
     <button
-      className={classNames(
-        "hover:bg-icon-gray-light focus:bg-icon-gray-dark group-focus:animate-pulse p-3 rounded-full grow-0 shrink-0 w-[2.8rem] h-[2.8rem] inline-flex items-center justify-center",
-        className
+      className={twMerge(
```
brampauwelyn, line 23: I see that twMerge merges classes without conflicts.
Is it something we want to have by default?
Maybe we could integrate it in the `classNames` utility function?
Or is it overkill

resources/ts/Components/IntegrationTypeCard.tsx
```diff
@@ -45,7 +48,11 @@ export const IntegrationTypeCard = ({
         <ul className="flex flex-col gap-3">
           {features.map((feature) => (
             <li key={feature} className="flex items-center gap-2">
-              <FontAwesomeIcon icon={faCheck} className="text-green-500" />
+              <FontAwesomeIcon
+                icon={faCheckSquare}
+                className="text-green-500"
+                size={"lg"}
```
brampauwelyn, line 54:
```suggestion
                size="lg"
```

brampauwelyn (approved):

resources/ts/Components/ButtonIcon.tsx
simon-debruijn: Until now we have been able to go without `twMerge` function just by using conditionals:

e.g.
Instead of 
`twMerge('mb-3 mb-6')`

We have just been doing
`condition ? 'mb-3' : 'mb-6'`

I'm not opposed to introducing `twMerge` if it is really needed. Then I would also integrate it in classNames.
But it is an extra runtime dependency we haven't needed until now.

resources/ts/Components/Card.tsx
```diff
@@ -105,16 +107,14 @@ export const Card = ({
                 {title}
               </Heading>
             )}
-
             {!!badge && (
-              <span className="bg-white text-publiq-blue border border-publiq-blue text-xs font-medium  mr-2 px-2.5 py-0.5 rounded">
-                {badge}
+              <span className=" text-publiq-gray-medium-dark bg-publiq-gray-light uppercase border border-publiq-gray-medium-dark text-xs font-medium  mr-2 px-2.5 py-0.5 rounded">
+                {badge.replace("-", " ")}
```
simon-debruijn, line 112: Is it only the first `-`? Otherwise you can use `replaceAll`

simon-debruijn (approved):

resources/ts/Components/ButtonIcon.tsx
Anahkiasen: To me the main benefit of `twMerge` is for passing extra classes multiple levels down. I do agree for most use cases, I prefer a behavioral prop on the components that then affect their appearance. But for a lot of one off style changes it's not always realistic without making the components API too bloated, so I like the merging approach as you can easily pass a margin or border class multiple levels down and have it override only the necessary classes.

But I understand the will to limit the number of runtime dependencies, I'd say PPF is ok on the basis that it's still a mainly back-end project with not that much front-end for now so we have some leeway.

PR #694, opened 2024-01-05
feature/PPF 250

Anahkiasen:
- Split button up
- Split URLs in configuration
- Move URL building to new class
- Use live platform if integration is active
- Nest configuration options
- Linting
- Add missing trailing slash in default
- Update logic in user-facing part too
- Remove mistake copy paste
- Show buttons conditionally

### DIFF
--- a/app/ProjectAanvraag/ProjectAanvraagServiceProvider.php
+++ b/app/ProjectAanvraag/ProjectAanvraagServiceProvider.php
@@ -25,7 +25,7 @@ final class ProjectAanvraagServiceProvider extends ServiceProvider
             return new ProjectAanvraagClient(
                 new Client(
                     [
-                        'base_uri' => config('project_aanvraag.base_uri'),
+                        'base_uri' => config('project_aanvraag.base_uri.live'),
                         'http_errors' => false,
                     ]
                 ),
--- a/resources/ts/Components/IntegrationCard.tsx
+++ b/resources/ts/Components/IntegrationCard.tsx
@@ -26,16 +26,6 @@ const OpenWidgetBuilderButton = ({ id, type }: any) => {
     return null;
   }
 
-  let url;
-  const widgetBuilderAppUrl = "";
-  if (ctrl.project.totalWidgets > 0) {
-    ctrl.actionButtonLinkLive = `${widgetBuilderAppUrl}/project/${id}`;
-    ctrl.actionButtonLinkTest = `${widgetBuilderAppUrl}/project/${id}`;
-  } else {
-    ctrl.actionButtonLinkLive = `${widgetBuilderAppUrl}/project/${id}/page/add`;
-    ctrl.actionButtonLinkTest = `${widgetBuilderAppUrl}/project/${id}/page/add`;
-  }
-
   return (
     <ButtonSecondary>
       <a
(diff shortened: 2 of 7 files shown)


app/Nova/Resources/Integration.php
```diff
@@ -178,7 +178,7 @@ public function actions(NovaRequest $request): array
                     return $model->status === IntegrationStatus::Draft->value;
                 }),
 
-            (new OpenWidgetBeheer())
+            (new OpenWidgetManager())
```
LucWollants, line 181: Is the label of the custom Nova action still "Open Widget Beheer" or did it become "Open Widget Manager"?

app/ProjectAanvraag/ProjectAanvraagServiceProvider.php
```diff
@@ -25,7 +25,7 @@ public function register(): void
             return new ProjectAanvraagClient(
                 new Client(
                     [
-                        'base_uri' => config('project_aanvraag.base_uri'),
+                        'base_uri' => config('project_aanvraag.base_uri.live'),
```
LucWollants, line 28: This becomes more tricky now, this is also dependent on the integration that gets created. So I think the HTTP client needs to be created inside the `createWidget` method depending on the Integration status.

LucWollants: Maybe the easiest approach is to extend the `CreateWidgetRequest` with the `IntegrationStatus`

LucWollants (changes requested):
Some extra changes are needed for `ProjectAanvraagClient`

app/Nova/Resources/Integration.php
Anahkiasen: I left if as Beheer yes

LucWollants: I was confused, the `$name` is defined inside the action itself and not here.

resources/ts/Components/IntegrationCard.tsx
```diff
@@ -20,6 +21,25 @@ const productTypeToPath = {
...
 
+const OpenWidgetBuilderButton = ({ id, type }: Pick<Props, "id" | "type">) => {
+  const { t } = useTranslation();
+  if (type !== "widgets") {
+    return null;
+  }
+
+  return (
+    <ButtonSecondary>
+      <a
```
simon-debruijn, line 32: It would be better to use a link styled as a secondary button.
`ButtonLinkSecondary` exists if I'm not mistaken

routes/web.php
```diff
@@ -99,5 +99,5 @@
     Route::post('/integrations/{id}/coupon', [IntegrationController::class, 'activateWithCoupon']);
     Route::post('/integrations/{id}/organization', [IntegrationController::class, 'activateWithOrganization']);
 
-    Route::get('/integrations/{id}/widget', [IntegrationController::class, 'showWidget']);
+    Route::get('/integrations/{integration}/widget', [IntegrationController::class, 'showWidget']);
```
simon-debruijn, line 102: What is this vodoo magic? 😛 Is Laravel able to automatically convert id to `integrationModel`?

simon-debruijn (approved):

routes/web.php
LucWollants: It is called Route Model Binding: https://laravel.com/docs/10.x/routing#route-model-binding

simon-debruijn: Nice 💪🏻

PR #715, opened 2024-01-15
Fix some handlers being named listeners

Anahkiasen:
### Fixed
- Fix some handlers being named listeners


LucWollants:
@Anahkiasen should the folder name `Jobs` also be changed to `Commands`?

Anahkiasen:
@LucWollants Not sure because while command is more correct, it's also confusing in a Laravel context since only CLI commands are refered as such with plain action commands being jobs. So one is closer to the actual pattern but this way is perhaps less confusing within the scope of Laravel's terminology, so not sure.

LucWollants (approved):

PR #718, opened 2024-01-16
UID-956 - Use acc for testing purposes

brampauwelyn:
### Changed
- Use acc for testing purposes

---
Ticket: https://jira.uitdatabank.be/browse/UID-956


resources/ts/Components/UitIdWidget.tsx
```diff
@@ -22,7 +22,7 @@ export const UitIdWidget = ({
   const widgetConfig = useMemo(
     () =>
       JSON.stringify({
-        $schema: "https://assets.uit.be/uitid-widget/config-schema.json",
+        $schema: `${import.meta.env.VITE_UITID_WIDGET_URL}config-schema.json`,
```
simon-debruijn, line 25: It would make more sense for me if the url from .env was /uitid-widget without trailing forward slash

simon-debruijn (approved):

PR #735, opened 2024-01-23
Use model binding for showWidget

simon-debruijn:
### Fixed

---
Ticket: https://jira.uitdatabank.be/browse/...


simon-debruijn:
This was to test if model binding would work on the tests, since it didn't work on my local machine, but would on Luc's local machine. This can be closed since it doesn't work in ci

PR #736, opened 2024-01-24
PPF-249-Implement icons and illustrations

vhande:
### Added
-Hero.svg
-HeroImage
-IconCustomizedSupport
-IconDocumentation
-IconReleaseNotes
-IconRoadmap
-IconSlack
-IconEntryApi
-IconSearchApi
-IconWidgets

localhost_nl_integraties_nieuw
localhost_nl

---
Ticket: https://jira.uitdatabank.be/browse/PPF-249


resources/ts/Components/Card.tsx
```diff
@@ -58,14 +58,7 @@ export const Card = ({
...
         )}
-        {img && (
-          <img
-            src={img}
-            className={classNames(
-              img && "h-full w-auto aspect-square max-h-[12rem] object-contain"
-            )}
-          ></img>
-        )}
+        {img && img}
```
brampauwelyn, line 61: Do we need the condition?
Can it not simply be?
```suggestion
        {img}
```

resources/ts/Components/HeroImage.tsx
```diff
@@ -0,0 +1,6 @@
+import React from "react";
+import Hero from "../../images/Hero.svg";
+
+export const HeroImage = () => {
+  return <img className="object-cover" src={Hero}></img>;
```
brampauwelyn, line 5:
```suggestion
  return <img className="object-cover" src={Hero} />;
```

brampauwelyn (approved):

resources/ts/Components/Card.tsx
vhande: Some instances of the card component do not include an image.

PR #789, opened 2024-02-29
PPF-38-Activation dialog

vhande:
### Added
-Activation Dialog

dialog

---
Ticket: https://jira.uitdatabank.be/browse/PPF-38


resources/ts/Components/StatusLight.tsx
```diff
@@ -33,34 +32,9 @@ export const StatusLight = ({ status }: Props) => {
...
-                </Link>
-              </>
-            )}
-          </div>
-          <ButtonPrimary className="self-start">
-            {t("integrations.status.activate")}
-          </ButtonPrimary>
-        </div>
+      {(status === "pending_approval_integration" ||
+        status === "pending_approval_payment") && (
```
brampauwelyn, line 36: You can also write something like this
```suggestion
      {["pending_approval_integration", "pending_approval_payment"].includes(status) && (
```

resources/ts/Components/ActivationDialog.tsx
```diff
@@ -0,0 +1,267 @@
...
+            <FormElement
+              label={`${t("integrations.activation_dialog.subscription_plan")}`}
+              component={
+                <p className="text-sm">
+                  {subscription &&
+                    `${subscription.category} (${
+                      subscription.currency === "EUR"
+                        ? "€"
+                        : subscription.currency
+                    } ${subscription.fee / 100})`}
```
brampauwelyn, line 101: Are these brackets necessary?
```suggestion
                    `${subscription.category} ${
                      subscription.currency === "EUR"
                        ? "€"
                        : subscription.currency
                    } ${subscription.fee / 100}`}
```

simon-debruijn (changes requested):

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -263,12 +274,15 @@ public function activateWithOrganization(string $id, CreateOrganizationRequest $
...
-        return Redirect::route(
-            TranslatedRoute::getTranslatedRouteName($request, 'integrations.show'),
-            [
-                'id' => $id,
-            ]
-        );
+        return Redirect::back();
+    }
+
+    public function distributeAuth0Clients(string $id): RedirectResponse
```
LucWollants, line 280: I don't think the front end needs to handle the distribution of a key. A key becoming distributed is a "side-effect" of activating an integration. When an integration is activated the production key gets set to distributed.

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -240,19 +249,21 @@ public function updateOrganization(string $id, UpdateOrganizationRequest $reques
...
-            TranslatedRoute::getTranslatedRouteName($request, 'integrations.show'),
-            [
-                'id' => $id,
-            ]
-        );
+            return Redirect::back();
+
+        } catch (ModelNotFoundException $exception) {
+            return Redirect::back()->withErrors([
+                'coupon' => 'Invalid coupon',
```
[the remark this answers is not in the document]
simon-debruijn: You now implemented the translation in the frontend. But laravel contains validation translations in the backend. I would prefer if 'Invalid coupon' gets translated directly.
This is not blocking the PR, we can change it after.

resources/translations/en.json
```diff
@@ -142,6 +142,16 @@
...
     },
+    "activation_dialog": {
+      "title": "Activatie aanvragen",
+      "subscription_plan": "Prijsplan",
+      "price": "Totaalbedrag",
+      "contact": "E-mail boekhouding",
+      "contact_description": "Dit adres ontvangt de betalinginstructies, -herinneringen en de factuur.",
+      "coupon": "Coupon",
+      "invalid_coupon": "Invalid coupon",
+      "alrady_used_coupon": "Coupon is already used"
```
simon-debruijn, line 153:
```suggestion
      "already_used_coupon": "Coupon is already used"
```

resources/ts/Components/ActivationDialog.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const setErrorMessage = useMemo(() => {
+    if (couponForm.errors.coupon) {
+      switch (couponForm.errors.coupon) {
+        case "Coupon is already used":
+          return t("integrations.activation_dialog.alredy_used_coupon");
+        default:
+          return t("integrations.activation_dialog.invalid_coupon");
+      }
+    }
+  }, [couponForm.errors, t]);
```
simon-debruijn, line 80:
```suggestion
  const errorMessage = useMemo(() => {
    if (!couponForm.errors.coupon) return;
    
    switch (couponForm.errors.coupon) {
      case "Coupon is already used":
          return t("integrations.activation_dialog.already_used_coupon");
      default:
        return t("integrations.activation_dialog.invalid_coupon");
    }
  }, [couponForm.errors, t]);
```

resources/ts/Components/ActivationFlow.tsx
```diff
@@ -0,0 +1,43 @@
...
+        onClose={() => {
+          router.get(url.toString(), {
+            isDialogVisible: undefined,
+          });
+          setIsActivationDialogVisible(false);
+        }}
+        id={id}
+        subscription={subscription}
+        type={type}
+        email={email!}
```
simon-debruijn, line 39:
```suggestion
        email={email}
```

If email is always defined, we better make it not optional in the props

resources/ts/Components/CopyText.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { ComponentProps, useRef, useState } from "react";
+import { Tooltip } from "./Tooltip";
+import { ButtonIconCopy } from "./ButtonIconCopy";
+import { useTranslation } from "react-i18next";
+
+export type Props = ComponentProps<"div">;
+
+export const CopyText = ({ children }: Props) => {
```
simon-debruijn, line 8: If it is only children where using as prop, you can better define the type like so:
```ts
type Props = { children: ReactNode }
```

resources/ts/Components/Integrations/Detail/CredentialsAuth0Clients.tsx
```diff
@@ -0,0 +1,159 @@
+import React, { useMemo } from "react";
+import { Heading } from "../../Heading";
+import { useTranslation } from "react-i18next";
+import {
+  Auth0Client,
+  Integration,
+  UiTiDv1Consumer,
+} from "../../../Pages/Integrations/Index";
+import { StatusLight } from "../../StatusLight";
+//import { ButtonPrimary } from "../../ButtonPrimary";
```
simon-debruijn, line 10:
```suggestion
```

resources/ts/Components/Integrations/Detail/CredentialsAuth0Clients.tsx
```diff
@@ -0,0 +1,159 @@
...
+      value: auth0ProdClient?.clientSecret,
+    },
+  ];
+
+  return (
+    <>
+      <Heading className="font-semibold lg:min-w-60" level={4}>
+        {t("details.credentials.uitid_v2")}
+      </Heading>
+      {/*{!auth0ProdClient ? (
```
simon-debruijn, line 73: Are you leaving this code, for when there is more clarity regarding the visibility of credentials?

resources/ts/Components/StatusLight.tsx
```diff
@@ -33,35 +32,9 @@ export const StatusLight = ({ status }: Props) => {
...
-                </Link>
-              </>
-            )}
-          </div>
-          <ButtonPrimary className="self-start">
-            {t("integrations.status.activate")}
-          </ButtonPrimary>
-        </div>
-      )}
+      {["pending_approval_integration", "pending_approval_payment"].includes(
```
simon-debruijn, line 35: It could be safer to use the `Status.PENDING_APPROVAL_...` constant here. Less brittle that just strings

simon-debruijn (changes requested):

resources/ts/Components/Integrations/Detail/CredentialsAuth0Clients.tsx
vhande: Yes

PR #808, opened 2024-03-08
Correct subscription plans

corneelwille:
### Changed
- tariffs of the subscription plans

---
Ticket: https://jira.publiq.be/browse/PPF-287

### DIFF
--- a/database/seeders/SubscriptionsSeeder.php
+++ b/database/seeders/SubscriptionsSeeder.php
@@ -38,7 +38,7 @@ final class SubscriptionsSeeder extends Seeder
                 SubscriptionCategory::Basic,
                 IntegrationType::SearchApi,
                 Currency::EUR,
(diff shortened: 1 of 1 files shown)


LucWollants:
I think something else is wrong because it is normal that `€125,00` gets stored as `12500` inside the database. This is done to avoid rounding errors. So somewhere in the presentation layer the divide by `100` is no longer done.

LucWollants:
> I think something else is wrong because it is normal that `€125,00` gets stored as `12500` inside the database. This is done to avoid rounding errors. So somewhere in the presentation layer the divide by `100` is no longer done.

I'm wrong, these are indeed the float values from the domain layer and these are float values.

simon-debruijn:
Ah you're right the multiplication is done in `SubscriptionModel` itself

PR #826, opened 2024-03-11
PPF-258 Expose distributed_at in Nova

LucWollants:
### Added
- Added `distributed_at` column in Nova on Auth0 Client and UiTiD Consumer
- Added action to distribute Auth0 Client or UiTiD Consumer

### Changed
- Show activate/block actions based on the state of the Auth0 Client or UiTiD Consumer
- Show actions on Integration correctly based on state and type

### Note
- This also fixes https://jira.publiq.be/browse/PPF-294

---
Ticket: https://jira.uitdatabank.be/browse/PPF-258

### DIFF
--- a/app/Domain/Integrations/Models/IntegrationModel.php
+++ b/app/Domain/Integrations/Models/IntegrationModel.php
@@ -51,6 +51,21 @@ final class IntegrationModel extends UuidModel
         'partner_status' => IntegrationPartnerStatus::THIRD_PARTY,
     ];
 
+    public function canBeActivated(): bool
+    {
+        return $this->status !== IntegrationStatus::Active->value;
+    }
+
+    public function canBeBlocked(): bool
+    {
+        return $this->status !== IntegrationStatus::Blocked->value;
+    }
+
+    public function isWidgets(): bool
+    {
+        return $this->type === IntegrationType::Widgets->value;
+    }
+
     protected static function booted(): void
     {
         self::created(
--- a/app/Nova/Resources/Auth0Client.php
+++ b/app/Nova/Resources/Auth0Client.php
@@ -7,11 +7,13 @@ namespace App\Nova\Resources;
 use App\Auth0\Auth0Tenant;
 use App\Auth0\CachedAuth0ClientGrants;
 use App\Auth0\Models\Auth0ClientModel;
+use App\Nova\ActionGuards\ActionGuard;
 use App\Nova\ActionGuards\Auth0\ActivateAuth0ClientGuard;
 use App\Nova\ActionGuards\Auth0\BlockAuth0ClientGuard;
 use App\Nova\Actions\Auth0\ActivateAuth0Client;
 use App\Nova\Actions\Auth0\BlockAuth0Client;
 use App\Nova\Resource;
+use Illuminate\Http\Request;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Log;
 use Laravel\Nova\Fields\DateTime;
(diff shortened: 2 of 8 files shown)


app/Domain/Integrations/Models/IntegrationModel.php
```diff
@@ -51,6 +51,21 @@ final class IntegrationModel extends UuidModel
         'partner_status' => IntegrationPartnerStatus::THIRD_PARTY,
     ];
 
+    public function canBeActivated(): bool
+    {
+        return $this->status !== IntegrationStatus::Active->value;
```
JonasVHG, line 56: `return $this->status !== IntegrationStatus::Active->value && $this->status !== IntegrationStatus::Blocked->value;`

If you block an `Integration` in `Draft`, you still see the option to `activate` it, which seems counterintuitive to me.

app/Nova/Resources/UiTiDv1.php
```diff
@@ -121,15 +121,15 @@ public function actions(NovaRequest $request): array
                 ->confirmText('Are you sure you want to activate this consumer?')
                 ->confirmButtonText('Activate')
                 ->cancelButtonText("Don't activate")
-                ->canSee(fn(Request $request) => $this->canActivate($request, $this->resource))
+                ->canSee(fn (Request $request) => $this->canActivate($request, $this->resource))
                 ->canRun(fn ($request, $model) => $this->canActivate($request, $model)),
```
JonasVHG, line 127: >

app/Domain/Integrations/Models/IntegrationModel.php
LucWollants: Do you mean that you can `Block` or `Activate` an integration that has the status `Draft`?

JonasVHG: Blocking a `Draft` is `OK`. But should it be possible to `Activate` an `Integration` that has been `Blocked`. 
I assume this shouldn't be possible.

LucWollants: I added it because you can `Block` an Integration by "mistake", so now you have the option to `Activate` it again.

JonasVHG: Oh, that might indeed be a good idea. Though, I wonder about the flow. If an Integration was activated with a `coupon` and accidentally blocked, you could not reactivate it with the same `coupon`.
But maybe this could be tackled in another ticket?

LucWollants: Indeed, very good point. Maybe this needs an extra tweak and needs to become a reactivate without the need to enter a coupon or organization details?

Anahkiasen (approved):
Still wish we had more common ground between V1/V2, it's a shame we need to reiterate so much of the logic with only slight changes when both models are at their core the same thing

app/Nova/Resources/Auth0Client.php
```diff
@@ -108,27 +115,51 @@ public function actions(NovaRequest $request): array
...
                 ->confirmText('Are you sure you want to activate this client?')
                 ->confirmButtonText('Activate')
                 ->cancelButtonText("Don't activate")
-                ->canRun(function ($request, $model) {
-                    /** @var ActivateAuth0ClientGuard $guard */
-                    $guard = App::make(ActivateAuth0ClientGuard::class);
-                    return $guard->canDo($model->toDomain());
-                }),
+                ->canSee(fn (Request $request) => $this->canActivate($request, $this->resource))
+                ->canRun(fn ($request, $model) => $this->canActivate($request, $model)),
```
simon-debruijn, line 123: You could also type the function arguments of the canRun?
There are multiple instances, where they're not typed

simon-debruijn (approved):

LucWollants:
> Still wish we had more common ground between V1/V2, it's a shame we need to reiterate so much of the logic with only slight changes when both models are at their core about the same thing

This a valid remark, but in the future the v2 integration which is an Auth0 integration will be replaced with a Keycloack integration. This new integration is still a bit unclear and may have other functionalities, so better to keep it separated.

Another thing we noticed from UDB3 where we created an Offer as a base class for both Places and Events is that in the end the code became more complex with if statements.

PR #828, opened 2024-03-12
Fixed colors of background and cards

Anahkiasen:
### Fixed
- Fixed colors of background and cards

---
Ticket: https://jira.publiq.be/browse/PPF-275


tailwind.config.js
```diff
@@ -35,7 +35,8 @@ module.exports = {
         "publiq-gray-dark": "hsl(0,0%,23%)",
         "publiq-gray-medium-dark": "hsl(0,0%,64%)",
         "publiq-gray-medium": "hsl(0,0%,94%)",
-        "publiq-gray-light": "hsl(0,0%,98%)",
+        "publiq-gray-light": "hsl(0,0%,96%)",
+        "publiq-gray-lighter": "hsl(0,0%,98%)",
```
Anahkiasen, line 39: Still not sure this scale makes any sense, perhaps we should have used numbers like Tailwind does between the jump from medium-dark to medium to light is certainly weird but would mean renaming a lot of classes :/

brampauwelyn (approved):

tailwind.config.js
simon-debruijn: Yes I am also for using something more scalable than light(er)(est) 😄

simon-debruijn (approved):

Anahkiasen:
@simon-debruijn Ok I converted the publiq gray to a Tailwind scale, in a way that the HSL of the previous colors matched the new ones, so it was pretty painless

PR #836, opened 2024-03-13
PPF-269-Hide credentials for widgets

vhande:
### Changed
- Hide credentials for widgets

widget1
widget2

---
Ticket: 
https://jira.uitdatabank.be/browse/PPF-269
https://jira.uitdatabank.be/browse/PPF-295

### DIFF
--- a/resources/ts/Components/ActivationRequest.tsx
+++ b/resources/ts/Components/ActivationRequest.tsx
@@ -7,7 +7,7 @@ import { useTranslateRoute } from "../hooks/useTranslateRoute";
 import { Integration } from "../Pages/Integrations/Index";
 import { IntegrationType } from "../types/IntegrationType";
 
-type Props = Integration;
+type Props = Pick<Integration, "id" | "type">;
 
 export const ActivationRequest = ({ id, type }: Props) => {
   const { t } = useTranslation();
--- a/resources/ts/Components/ActivationRequest.tsx
+++ b/resources/ts/Components/ActivationRequest.tsx
@@ -5,10 +5,12 @@ import { ButtonPrimary } from "./ButtonPrimary";
 import { router } from "@inertiajs/react";
 import { useTranslateRoute } from "../hooks/useTranslateRoute";
 import { Integration } from "../Pages/Integrations/Index";
+import { IntegrationType } from "../types/IntegrationType";
+import { Values } from "../types/Values";
 
-type Props = { id: Integration["id"] };
+type Props = { id: Integration["id"]; type: Values<typeof IntegrationType> };
 
-export const ActivationRequest = ({ id }: Props) => {
+export const ActivationRequest = ({ id, type }: Props) => {
   const { t } = useTranslation();
 
   const translateRoute = useTranslateRoute();
@@ -20,19 +22,21 @@ export const ActivationRequest = ({ id }: Props) => {
   return (
     <div className="flex flex-col gap-3">
       <div>
-        <Trans
-          i18nKey="integrations.status.pending_approval_integration_description"
-          t={t}
-          components={[
-            <Link
-              key={t(
-                "integrations.status.pending_approval_integration_description"
-              )}
-              href={t("integrations.status.before_going_live_link")}
-              className="text-publiq-blue-dark hover:underline"
-            />,
-          ]}
-        />
+        {type === IntegrationType.EntryApi && (
+          <Trans
+            i18nKey="integrations.status.pending_approval_integration_description"
+            t={t}
+            components={[
+              <Link
+                key={t(
+                  "integrations.status.pending_approval_integration_description"
+                )}
+                href={t("integrations.status.before_going_live_link")}
+                className="text-publiq-blue-dark hover:underline"
+              />,
+            ]}
+          />
+        )}
       </div>
       <ButtonPrimary className="self-start" onClick={handleActivate}>
         {t("integrations.status.activate")}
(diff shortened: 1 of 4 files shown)


resources/ts/Components/IntegrationCard.tsx
```diff
@@ -38,17 +39,63 @@ export const OpenWidgetBuilderButton = ({
...
+  status,
+  credentials,
+  onEdit,
+}: Props) => {
   const { t } = useTranslation();
 
   const integrationTypesInfo = useIntegrationTypesInfo();
 
+  const auth0TestClient = credentials.auth0.find(
+    (client) => client.integrationId === id && client.tenant === "test"
```
brampauwelyn, line 65: Maybe it's interesting to make a const with the different environments?
In case you need it in other places?  Or are the tenants already correctly typed?

Something like:

```typescript
const Environments = {
 Test: 'test',
 Acceptance: 'acc',
 Production: 'prod'
} as const;
```

simon-debruijn: Yes that constant already exists. and the type of tenant is also correct based on the constant.
So both are correct, it's a matter of preference using literal strings or the constant

resources/ts/Pages/Integrations/Index.tsx
```diff
@@ -98,9 +103,11 @@ export type Integration = {
 
 type Props = {
   integrations: Integration[];
-} & PaginationInfo;
+  credentials: Credentials;
+} & PaginationInfo &
+  Credentials;
```
Anahkiasen, line 108: This seems weird I'm assuming the last `& Credentials` can be removed perhaps?

resources/ts/Components/IntegrationCard.tsx
```diff
@@ -107,6 +248,11 @@ export const IntegrationCard = ({ id, name, type, status, onEdit }: Props) => {
                 product: t(`integrations.products.${type}`),
               })}
             </Link>
+            {type === "entry-api" && (
```
Anahkiasen, line 251: Would perhaps use the `IntegrationType` constant here

resources/ts/Components/ActivationRequest.tsx
```diff
@@ -5,10 +5,12 @@ import { ButtonPrimary } from "./ButtonPrimary";
 import { router } from "@inertiajs/react";
 import { useTranslateRoute } from "../hooks/useTranslateRoute";
 import { Integration } from "../Pages/Integrations/Index";
+import { IntegrationType } from "../types/IntegrationType";
+import { Values } from "../types/Values";
 
-type Props = { id: Integration["id"] };
+type Props = { id: Integration["id"]; type: Values<typeof IntegrationType> };
```
Anahkiasen, line 11: Think this could maybe be `type Props = Integration`

resources/ts/Components/Integrations/Detail/CredentialsUitidv1Consumers.tsx
```diff
@@ -47,37 +49,59 @@ export const CredentialsUitidv1Consumers = ({
...
-            <Alert variant="info">{t("details.credentials.info")}</Alert>
-          </div>
+            </>
+          )}
+        <div className="flex flex-col gap-2">
+          <Heading className="font-semibold" level={4}>
+            {t("details.credentials.live")}
+          </Heading>
+          <StatusLight status={status} />
+          {uiTiDv1ProdConsumer && (
```
Anahkiasen, line 79: Wondering if this condition should be `status === Active` since both conditions within rely on it and the to should be equivalent to me. Then you can simplify the condition within the div to a ternary on `type === Widgets`

resources/ts/Components/IntegrationCard.tsx
```diff
@@ -38,17 +41,69 @@ export const OpenWidgetBuilderButton = ({
...
+      client.integrationId === id && client.tenant === Auth0Tenant.Production
+  );
+  const uiTiDv1TestConsumer = credentials.uitidV1.find(
+    (client) =>
+      client.integrationId === id &&
+      client.environment === UiTiDv1Environment.Testing
+  );
+  const uiTiDv1ProdConsumer = credentials.uitidV1.find(
+    (client) =>
+      client.integrationId === id &&
```
Anahkiasen, line 81: I would perhaps only pass the credentials of the given integration to the integration card, it would make all those simpler and would mean the card doesn't have to worry about receiving another integration's data

Anahkiasen (changes requested):

simon-debruijn (approved):

resources/ts/Pages/Integrations/Index.tsx
```diff
@@ -168,17 +198,22 @@ const Index = ({ integrations, paginationInfo, credentials }: Props) => {
...
-          ))}
+          {integrationsWithCredentials.map(
+            ({ credentials, ...integration }) => (
+              <li className="flex w-full" key={integration.id}>
+                <IntegrationCard
+                  {...integration}
+                  auth0TestClient={credentials.auth0TestClient}
+                  auth0ProdClient={credentials.auth0ProdClient}
+                  uiTiDv1TestConsumer={credentials.uiTiDv1TestConsumer}
+                  uiTiDv1ProdConsumer={credentials.uiTiDv1ProdConsumer}
```
Anahkiasen, line 209: Would maybe keep those in an object, you can ensure the entries in it are correct through the `Props` of the `IntegrationCard` and it would avoid having to destructure -> restructure the integration.

Anahkiasen (approved):
Possible change but otherwise all good. I'm still partial to renaming the individual credentials to something that relies less on the implementation detail, like use V1/V2 and Test/Prod instead of auth0/uitidv1 but depends what @simon-debruijn and @brampauwelyn think since we still mention those implementation names in the copy, but I do feel it could make the code more straightfoward so not sure

simon-debruijn (approved):

PR #838, opened 2024-03-14
PPF-282 Add CLI command to migrate extra contacts from Insightly

LucWollants:
### Added
- Added CLI command to migrate extra contacts from Insightly

---
Ticket: https://jira.uitdatabank.be/browse/PPF-282

### DIFF
--- a/app/Console/Commands/Migrations/MigrateContacts.php
+++ b/app/Console/Commands/Migrations/MigrateContacts.php
@@ -72,7 +72,7 @@ final class MigrateContacts extends Command
             $this->info('---');
         }
 
-        return 1;
+        return 0;
     }
 
     private function migrateContact(UuidInterface $contactId, MigrationContact $migrationContact): void
--- a/app/Console/Commands/Migrations/MigrateContacts.php
+++ b/app/Console/Commands/Migrations/MigrateContacts.php
@@ -43,8 +43,7 @@ final class MigrateContacts extends Command
         $this->info('Migrating contacts...');
 
         $contacts = $this->readCsvFile('database/project-aanvraag/contacts.csv');
-        $migrationContacts = array_map(fn (array $contact) => new MigrationContact($contact), $contacts);
-        $contactsCount = count($migrationContacts);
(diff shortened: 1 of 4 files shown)


JonasVHG (approved):

app/Console/Commands/Migrations/MigrateContacts.php
```diff
@@ -0,0 +1,127 @@
...
+        $this->info('Migrating contacts...');
+
+        $contacts = $this->readCsvFile('database/project-aanvraag/contacts.csv');
+        $migrationContacts = array_map(fn (array $contact) => new MigrationContact($contact), $contacts);
+        $contactsCount = count($migrationContacts);
+
+        if ($contactsCount <= 0) {
+            $this->warn('No contacts to import');
+            return 0;
+        }
```
simon-debruijn, line 52: This can happen after the `$contacts = $this->readCsvFile` line. You can count the $contacts from the csv before the mapping

app/Console/Commands/Migrations/MigrateContacts.php
```diff
@@ -0,0 +1,127 @@
...
+                $this->warn($contactId . ' - Contact ' . $migrationContact->email() . ' already exists');
+            } catch (Exception $e) {
+                $this->error($contactId . ' - Failed importing contact ' . $migrationContact->email() . ': ' . $e->getMessage());
+            }
+
+            $this->info($contactId . ' - Ended importing contact ' . $migrationContact->email());
+            $this->info('---');
+        }
+
+        return 1;
```
simon-debruijn, line 75: Why is an error code (not 0) returned when the contacts have been migrated?
Or is it 0 (false) and 1 (true)?

simon-debruijn (approved):

app/Console/Commands/Migrations/MigrateContacts.php
LucWollants: This is indeed wrong, because `0` means `success`

simon-debruijn (approved):

PR #871, opened 2024-03-21
PPF-314 Create Nova action for activation

LucWollants:
### Changed
- Changed the activation to always require an organization and make the coupon optional. This new approach is already used for the Laravel Nova Activation action
- Marked several classes as deprecated for removal after frontend changes

### Deleted
- Deleted the existing activation action with only organization and the existing activation action with coupon only

---
Ticket: https://jira.uitdatabank.be/browse/PPF-314


grubolsch (approved):

app/Domain/Integrations/Events/IntegrationActivatedWithCoupon.php
```diff
@@ -7,6 +7,7 @@
 use Illuminate\Foundation\Events\Dispatchable;
 use Ramsey\Uuid\UuidInterface;
 
+// @deprecated
```
Anahkiasen, line 10: I think the @deprecated tags need to be in full PHPDocs to be properly recognized

Screenshot 2024-03-21 at 14 24 04

LucWollants: Good to know! Luckily the next step is to manually remove all that `@deprecated` code, so I will not change it 😏

PR #877, opened 2024-03-22
PPF-322 newsletter api

JonasVHG:
### Added

- `composer.json`: Added `mailjet/mailjet-apiv3-php`
-  `config/newsletter`: To read the config values for the `Mailjet`-api
- `routes/api`: Added `Route` to subscribe to newsletter
- `NewsletterController`: To add a `contact` to the `Mailjet`-api

### Related PR

- https://github.com/cultuurnet/appconfig/pull/580

---
Ticket: https://jira.publiq.be/browse/PPF-322


config/newsletter.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+return [
+    'apiKey' => env('NEWSLETTER_API_KEY'),
```
LucWollants, line 6: The default Laravel convention would be `api_key`, but I see we already have a mix.

app/Domain/Newsletter/Controllers/NewsletterController.php
```diff
@@ -0,0 +1,32 @@
...
+use App\Http\Controllers\Controller;
+use Mailjet\Client;
+use Mailjet\Resources;
+use Mailjet\Response;
+
+final class NewsletterController extends Controller
+{
+    public function handle(string $email): Response
+    {
+        $mailingListId = config('newsletter.id');
```
LucWollants, line 16: I guess there is no need for local variables.

app/Domain/Newsletter/Controllers/NewsletterController.php
```diff
@@ -0,0 +1,32 @@
...
+namespace App\Domain\Newsletter\Controllers;
+
+use App\Http\Controllers\Controller;
+use Mailjet\Client;
+use Mailjet\Resources;
+use Mailjet\Response;
+
+final class NewsletterController extends Controller
+{
+    public function handle(string $email): Response
```
LucWollants, line 14: Should the email address be validated?

app/Domain/Newsletter/Controllers/NewsletterController.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $mailingListId = config('newsletter.id');
+        $apiKey = config('newsletter.apiKey');
+        $apiSecret = config('newsletter.apiSecret');
+
+        $mailJetClient = new Client($apiKey, $apiSecret);
+        return $mailJetClient->post(
+            Resources::$ContactslistManagecontact,
```
LucWollants, line 22: This is weird 😅

config/newsletter.php
JonasVHG: Force of habit, I changed it, because I don't want to add it to the mix.

app/Domain/Newsletter/Controllers/NewsletterController.php
JonasVHG: I just followed the Mailjet Docs. I guess there is little we can do about it.

LucWollants: Indeed we can't change that. I should have commented and already resolved the conversation 🙂

app/Domain/Newsletter/Controllers/NewsletterController.php
```diff
@@ -0,0 +1,34 @@
...
+use App\Http\Controllers\Controller;
+use Mailjet\Client;
+use Mailjet\Resources;
+use Mailjet\Response;
+
+final class NewsletterController extends Controller
+{
+    public function handle(string $email): Response
+    {
+        if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
```
LucWollants, line 16: Maybe use the Laravel rules in this context? Something like this

```
$validator = Validator::make(
    ['email' => $email],
    ['email' => 'required|email']
);
```

app/Domain/Newsletter/Controllers/NewsletterController.php
LucWollants: Maybe use the Laravel rules in this context? Something like this

```
$validator = Validator::make(
    ['email' => $email],
    ['email' => 'required|email']
);
```

LucWollants: The frontend can even use the result of this validation, this is also done so for other forms.

app/Domain/Newsletter/Controllers/NewsletterController.php
```diff
@@ -5,22 +5,30 @@
...
-    public function handle(string $email): Response
+    public function handle(string $email): JsonResponse
     {
-        if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
-            throw new \InvalidArgumentException('Invalid email address');
+        $validator = Validator::make(
+            ['email' => $email],
+            ['email' => 'required|email']
+        );
+        if ($validator->fails()) {
```
LucWollants, line 21: I think you can do `$validator->validate();`

JonasVHG: But that throws a `ValidationException` which you need to catch?

LucWollants: I think that is also how the `rules` method works on `FormRequest`

JonasVHG: Indeed, there doesn't really seem an elegant solution to do this without a `redirect`.

LucWollants (approved):

PR #880, opened 2024-03-22
PPF-307 Remove distributed_at concept for clients and consumers

LucWollants:
### Removed
- Removed `distributed_at` concept for clients and consumers

### Note
- For now, the UI shows all keys, this will be fixed with https://jira.publiq.be/browse/PPF-306 where a visibility field will be added to the integration. This field can have values `v1`, `v2`, and `all`. This field in combination with the status of an integration controls the visibility of consumers and clients

---
Ticket: https://jira.uitdatabank.be/browse/PPF-307


Anahkiasen (approved):
you always think you're gonna have more time with your lines of code and then one day they just get deleted from you and thats where you learn life's harshest lesson :trollface:

PR #912, opened 2024-03-27
PPF-282 Create default v1 contacts

LucWollants:
### Changed
- Create default v1 contacts when migrating

### Removed
- Remove duplicated `RefreshDatabase` which is set on `TestCase`

---
Ticket: https://jira.uitdatabank.be/browse/PPF-282


Anahkiasen (approved):

tests/TestCase.php
```diff
@@ -4,13 +4,15 @@
...
 
+use Illuminate\Foundation\Testing\RefreshDatabase;
 use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
 use Illuminate\Support\Facades\Bus;
 use Illuminate\Support\Facades\Event;
 
 abstract class TestCase extends BaseTestCase
 {
     use CreatesApplication;
+    use RefreshDatabase;
```
simon-debruijn, line 15: I think I specifically didn't add `RefreshDatabase` to the base test case, because there are tests like the Mappers that don't need the database connection at all. The interaction with the database adds a overhead when wanting to run tests that don't use it.

simon-debruijn (changes requested):

tests/TestCase.php
LucWollants: Good point! I will revert it again.

simon-debruijn (approved):

PR #918, opened 2024-03-28
PPF-338 Set contact type on new relation type field

LucWollants:
### Changed
- Store the contact type on the new custom field relation type

---
Ticket: https://jira.uitdatabank.be/browse/PPF-338


LucWollants:
Only the names of the roles were changed to make it more clear inside Insightly.

PR #919, opened 2024-03-28
PPF-303-Hide urls when an integration only has v1 credentials

vhande:
### Changed
- Hide urls when an integration only has v1 credentials

---
Ticket: https://jira.uitdatabank.be/browse/PPF-303


resources/ts/Pages/Integrations/Index.tsx
```diff
@@ -22,6 +22,7 @@ import { IntegrationType } from "../../types/IntegrationType";
 import { Values } from "../../types/Values";
 import { Auth0Tenant } from "../../types/Auth0Tenant";
 import { UiTiDv1Environment } from "../../types/UiTiDv1Environment";
+import { KeyVisibility } from "../../types/KeyVisibility";
```
brampauwelyn, line 25: I think here you want to import the type?
```suggestion
import type { KeyVisibility } from "../../types/KeyVisibility";
```

brampauwelyn (approved):

resources/ts/Pages/Integrations/Index.tsx
simon-debruijn: I will create a PR to add this eslitn rule.
This way we don't have to worry about it.

https://typescript-eslint.io/rules/consistent-type-imports/

PR #926, opened 2024-03-29
PPF-370-Fix language toggle on integration detail page

vhande:
### Fixed
- Language toggle on integration detail page
---
Ticket: https://jira.uitdatabank.be/browse/PPF-370

### DIFF
--- a/resources/ts/Components/UitIdWidget.tsx
+++ b/resources/ts/Components/UitIdWidget.tsx
@@ -3,6 +3,7 @@ import { classNames } from "../utils/classNames";
 import { useTranslation } from "react-i18next";
 import { usePage } from "@inertiajs/react";
 import { useTranslateRoute } from "../hooks/useTranslateRoute";
+import { Integration } from "../Pages/Integrations/Index";
 
 export type WidgetConfigVariables = {
   profileUrl: string;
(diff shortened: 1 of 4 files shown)


resources/ts/Components/UitIdWidget.tsx
```diff
@@ -88,7 +92,7 @@ export const UitIdWidget = ({
             ></div>
             <a
               className={classNames(i18n.language === "en" && "active")}
-              href={`${translateRoute(`/${currentPage}`, "en")}?setLocale=true`}
+              href={`${translateRoute(`/${currentPage}`, "en", id)}?setLocale=true`}
```
simon-debruijn, line 95: Nice solution ⭐ We better make it a bit more generic, since we could have other urls with dynamic parts of the url.
The id passed here can be an object with variables, depending on the page. Then we have to create a mapping between the page (e.g. /detail) and the needed variables (id in this case). Something like:

```ts
const currentPageToVariables = {
    'integrations/detail': props.integration.id
    // other pages come here
}
```

simon-debruijn (changes requested):

resources/ts/Components/UitIdWidget.tsx
```diff
@@ -63,6 +63,12 @@ export const UitIdWidget = ({
     [component]
   );
 
+  const currentPageToVariables = {
+    "integrations/detail": {
+      id: props.integration?.id,
+    },
+  };
+
```
simon-debruijn, line 71:
```suggestion
  const currentPageToVariables: { [key: string]: unknown } = {
    [key: string]: unknown;
    
    "integrations/detail": {
      id: props.integration?.id,
    },
  };

```

This will satisfy the eslint error

simon-debruijn (approved):

PR #945, opened 2024-04-03
Add environment variable pointing to application for e2e tests

willaerk:


Jenkinsfile
```diff
@@ -92,6 +92,9 @@ pipeline {
 
         stage('Acceptance tests') {
             agent { label 'ubuntu && 20.04 && nodejs18' }
+            environment {
+                E2E_TEST_BASE_URL = 'https://platform-acc.publiq.be'
```
brampauwelyn, line 96: Normally we use a trailing slash.
But maybe it will also work like this.

```suggestion
                E2E_TEST_BASE_URL = 'https://platform-acc.publiq.be/'
```

brampauwelyn (approved):

brampauwelyn (approved):

PR #965, opened 2024-04-09
PPF-379 - Check nullability of coupon

simon-debruijn:
### Changed
- [Check nullability of coupon](https://github.com/cultuurnet/publiq-platform/commit/f9c2fbded074fda3c0fcb01a0bfd8f17320b5ec2)

### Fixed
- Billing info error on coupon

---
Ticket: https://jira.uitdatabank.be/browse/PPF-379

### DIFF
--- a/resources/ts/Components/Integrations/Detail/BillingInfo.tsx
+++ b/resources/ts/Components/Integrations/Detail/BillingInfo.tsx
@@ -50,7 +50,7 @@ export const BillingInfo = ({
             />
           }
         />
-        {coupon && coupon.isDistributed && (
+        {coupon?.isDistributed && (
           <Alert
             className={"col-span-2 col-start-2"}
             variant="success"
--- a/resources/ts/Components/Integrations/Detail/BillingInfo.tsx
+++ b/resources/ts/Components/Integrations/Detail/BillingInfo.tsx
@@ -75,19 +75,17 @@ export const BillingInfo = ({
           <Heading level={4} className="font-semibold">
             {t("details.billing_info.to_pay")}
           </Heading>
-          {coupon && (
-            <div className="w-full block relative md:min-w-[40rem]">
-              {t(`pricing_plan.basic.price`, {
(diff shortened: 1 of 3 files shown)


vhande (approved):

resources/ts/Components/Integrations/Detail/BillingInfo.tsx
```diff
@@ -75,17 +75,19 @@ export const BillingInfo = ({
...
-              price: formatPricing({
-                currency: subscription.currency,
-                price: Math.max(
-                  subscription.fee / 100 - (coupon?.reduction ?? 0),
-                  0
-                ),
-              }),
-            })}
-          </div>
+          {coupon && (
```
Anahkiasen, line 78: Wouldn't that condition hide the price even in case of no coupon?

resources/ts/Components/Integrations/Detail/BillingInfo.tsx
```diff
@@ -50,7 +50,7 @@ export const BillingInfo = ({
             />
           }
         />
-        {coupon.isDistributed && (
+        {coupon && coupon.isDistributed && (
```
Anahkiasen, line 53: Would `?.` here maybe :trollface:

resources/ts/types/Integration.ts
```diff
@@ -22,7 +22,7 @@ export type Integration = {
   name: string;
   description: string;
   subscriptionId: string;
-  coupon: Coupon;
+  coupon?: Coupon | null;
```
Anahkiasen, line 25: I feel the first `?` should not be here since the property will always be there but might be null whereas here it can be both missing and null

resources/ts/Components/Integrations/Detail/BillingInfo.tsx
simon-debruijn: Yes this isn't needed, nice catch

PR #981, opened 2024-04-11
PPF-389 Expose key visibility upgrade

LucWollants:
### Changed
- Exposed key visibility upgrade on Integration

---
Ticket: https://jira.uitdatabank.be/browse/PPF-389

### DIFF
--- a/app/Domain/KeyVisibilityUpgrades/KeyVisibilityUpgrade.php
+++ b/app/Domain/KeyVisibilityUpgrades/KeyVisibilityUpgrade.php
@@ -8,14 +8,14 @@ use App\Domain\Integrations\KeyVisibility;
 use DateTimeImmutable;
 use Ramsey\Uuid\UuidInterface;
 
-final readonly class KeyVisibilityUpgrade
+final class KeyVisibilityUpgrade
 {
     private DateTimeImmutable $createdAt;
 
     public function __construct(
-        public UuidInterface $id,
-        public UuidInterface $integrationId,
-        public KeyVisibility $keyVisibility
+        public readonly UuidInterface $id,
+        public readonly UuidInterface $integrationId,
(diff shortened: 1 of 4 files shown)


app/Domain/KeyVisibilityUpgrades/KeyVisibilityUpgrade.php
```diff
@@ -5,14 +5,29 @@
 namespace App\Domain\KeyVisibilityUpgrades;
 
 use App\Domain\Integrations\KeyVisibility;
+use DateTimeImmutable;
 use Ramsey\Uuid\UuidInterface;
 
-final readonly class KeyVisibilityUpgrade
+final class KeyVisibilityUpgrade
 {
+    private DateTimeImmutable $createdAt;
```
simon-debruijn, line 13: Can't this be included as readonly in the constructor?
`createdAt` seems to be required (not nullable)

simon-debruijn (changes requested):

app/Domain/KeyVisibilityUpgrades/KeyVisibilityUpgrade.php
LucWollants: I did hide it from the constructor, because when requesting an upgrade the `createdAt` is not needed, it is controlled by saving it inside the database.

simon-debruijn: Than it makes sense to make `createdAt` `?DateTimeImmutable` right?

Otherwise you can do

```php
$upgrade = new KeyVisibilityUpgrade(Uuid::uuid4(), Uuid::uuid4(), KeyVisibility::all);
$upgrade->getCreatedAt(); // this won't return a DateTimeImmutable
```

simon-debruijn (approved):

PR #982, opened 2024-04-11
PPF-376-Fix key visibility upgrade

vhande:
### Fixed
-Key visibility upgrade
---
Ticket: https://jira.uitdatabank.be/browse/PPF-376

### DIFF
--- a/resources/ts/Components/Integrations/Detail/CredentialsAuthClients.tsx
+++ b/resources/ts/Components/Integrations/Detail/CredentialsAuthClients.tsx
@@ -53,7 +53,7 @@ export const CredentialsAuthClients = ({
 
   const handleKeyVisibilityUpgrade = () =>
     router.post(`/integrations/${id}/upgrade`, {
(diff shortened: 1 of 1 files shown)


simon-debruijn (approved):

resources/ts/Components/Integrations/Detail/CredentialsAuthClients.tsx
```diff
@@ -52,7 +52,9 @@ export const CredentialsAuthClients = ({
   ];
 
   const handleKeyVisibilityUpgrade = () =>
-    router.post(`/integrations/${id}/upgrade`, { keyVisibility });
+    router.post(`/integrations/${id}/upgrade`, {
+      keyVisibility: KeyVisibility.all,
```
LucWollants, line 56: This needs to be `v2`  because you request an upgrade to `v2`. It is true this will result in a key visibility of `all` on the integration, but the upgrade request is from `v1` to `v2` so in this context it needs to be `v2`

PR #985, opened 2024-04-11
PPF-389-Fix expose key visibility upgrade

vhande:
### Fixed
-  expose key visibility upgrade
---
Ticket: https://jira.uitdatabank.be/browse/PPF-389


app/Domain/KeyVisibilityUpgrades/KeyVisibilityUpgrade.php
```diff
@@ -31,4 +31,27 @@ public function getCreatedAt(): ?DateTimeImmutable
...
         return $this->createdAt;
     }
+
+    private function getDaysLeft(): int
+    {
+        if ($this->createdAt === null) {
+            throw new \Exception('Property createdAt should exist');
+        }
+
+        $days = $this->createdAt->diff(new DateTimeImmutable())->days;
```
LucWollants, line 41: I think this is the days passed since the creation, not the days until the old key expires.

LucWollants: To be clear the date of expiration is not stored, because that would be tricky when the grace period changes. So an easy work around is to just store the creation date of the upgrade request and use the controller or UI to take into account the grace period.

resources/ts/Components/Integrations/Detail/CredentialsLegacyAuthConsumers.tsx
```diff
@@ -25,8 +32,16 @@ export const CredentialsLegacyAuthConsumers = ({
   subscription,
   type,
   keyVisibility,
+  oldCredentialsExpirationDate,
```
simon-debruijn, line 35:
```suggestion
  oldCredentialsExpirationDate: oldCredentialsExpirationDateString,
```

resources/ts/Components/Integrations/Detail/CredentialsLegacyAuthConsumers.tsx
```diff
@@ -25,8 +32,16 @@ export const CredentialsLegacyAuthConsumers = ({
...
 }: Props) => {
-  const { t } = useTranslation();
+  const { t, i18n } = useTranslation();
+  const legacyCredentialsExpirationDate = new Date(
+    oldCredentialsExpirationDate
+  );
```
simon-debruijn, line 40:
```suggestion
  const oldCredentialsExpirationDate = new Date(
    oldCredentialsExpirationDateString
  );
```

resources/ts/Components/Integrations/Detail/CredentialsLegacyAuthConsumers.tsx
```diff
@@ -25,8 +32,16 @@ export const CredentialsLegacyAuthConsumers = ({
...
+  const timeLeft = formatDistanceToNow(legacyCredentialsExpirationDate, {
+    locale: languageToLocale[i18n.language],
+  });
```
simon-debruijn, line 43:
```suggestion
  const timeLeft = formatDistanceToNow(oldCredentialsExpirationDate, {
    locale: languageToLocale[i18n.language],
  });
```

simon-debruijn (approved):

PR #1001, opened 2024-04-15
PPF-386-Handle 500 error

vhande:
### Fixed
- Handle 500 error
---
Ticket: https://jira.uitdatabank.be/browse/PPF-386


app/Exceptions/Handler.php
```diff
@@ -61,14 +61,14 @@ public function register(): void
...
-                return $this->handle500Error($exception);
-            }
 
+        $this->renderable(function (Exception $exception) {
             if ($exception instanceof HttpException && $exception->getStatusCode() === 404) {
                 return $this->handle404Error($exception);
             }
+            if (!$exception instanceof HttpException) {
+                return $this->handle500Error($exception);
+            }
```
simon-debruijn, line 71:
```suggestion
        $this->renderable(function (Exception $exception) {
            if (!$exception instanceof HttpException) {
                return $this->handle500Error($exception);
            }
            if ($exception->getStatusCode() === 404) {
                return $this->handle404Error($exception);
            }
            if ($exception->getStatusCode() >= 500) {
                return $this->handle500Error($exception);
            }            
```

We still have to check http exceptions that are 500.

simon-debruijn (changes requested):

simon-debruijn (approved):

PR #1003, opened 2024-04-15
PPF-396-Handle duplicate contact error

vhande:
### Added
- Handle duplicate contact error

---
Ticket: https://jira.uitdatabank.be/browse/PPF-396


app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -198,12 +199,15 @@ public function updateUrls(UpdateIntegrationUrlsRequest $request, string $id): R
...
-
-        DB::transaction(function () use ($contacts) {
-            foreach ($contacts as $contact) {
-                $this->contactRepository->save($contact);
-            }
-        });
+        try {
+            DB::transaction(function () use ($contacts) {
+                foreach ($contacts as $contact) {
+                    $this->contactRepository->save($contact);
```
LucWollants, line 205: Another option is to catch the exception here and continue the loop. If we do that the resulting user experience is maybe better.

simon-debruijn: This should be possible. But then we will have to build up the errors depending on which contacts failed and show map them correctly in the frontend.
Now it's fail or success. With your proposed change it will be more selective.
@vhande Let us know if you need help implementing this, if you plan to

simon-debruijn (changes requested):

app/Domain/Integrations/Controllers/IntegrationController.php
LucWollants: You even don't have to show an error if the contact was already part of the integration, it will be a "no operation" case

simon-debruijn: That is true, but what if they use a general email `dev@publiq.be` and try to add another contact with the same email but different name. We probably should warn them about the fact they cannot add another contact with the same email? Otherwise it will look successful, but no data has been added

LucWollants: Valid point, I was only thinking about duplicates inside the list of contacts.

vhande: @LucWollants I would like to ask you if it's really necessary, since the user can only edit/delete 1 contact at a time, so it's 1 request per time, and it can fail only once anyway.

simon-debruijn (approved):
This PR does handle the duplicate contact exception, let's merge this. We still have to refactor contacts update logic, since it is flawed. We'll do it in a separate PR.

app/Domain/Integrations/Controllers/IntegrationController.php
simon-debruijn: The UI itself only sends 1 updated contact at the time, but the endpoint supports multiple.
We better refactor this to just have a updateContact singular, deleteContact, addContact.

PR #1063, opened 2024-05-03
PPF-435 fetching token keycloak

grubolsch:
### Added
- Added a class to fetch a token using client_credentials for Keycloak.
- I created an interface TokenStrategy to support later token-fetching-strategies like username/password fetching if required.
- Added config files to configure said Keycloak

This does not add anything visibile in the UI yet!

---


app/Keycloak/TokenStrategy/ClientCredentials.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function __construct(
+        private readonly ClientInterface $client,
+        private readonly Config $config
+    ) {
+    }
+
+    public function fetchToken(Realm $realm): string
+    {
+        if ($this->accessToken === null) { //lazy loading
```
simon-debruijn, line 30: this can become an early return instead

tests/Keycloak/TokenStrategy/ClientCredentialsTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->config = new Config(
+            true,
+            'https://keycloak.example.com/',
+            'php_client',
+            'a_true_secret',
+            new Realm('uitidpoc', 'Acceptance')
+        );
+    }
+
+    public function testHappyPath(): void
```
simon-debruijn, line 32: Tests are snake_cased

tests/Keycloak/TokenStrategy/ClientCredentialsTest.php
```diff
@@ -0,0 +1,92 @@
...
+        ]);
+
+        $handlerStack = HandlerStack::create($mock);
+        $clientCredentials = new ClientCredentials(new Client(['handler' => $handlerStack]), $this->config);
+        $token = $clientCredentials->fetchToken(Realm::getMasterRealm());
+
+        $this->assertEquals('pqeaefosdfhbsdq', $token);
+    }
+
+    public function testFailedBecauseIntegrationIsDisabled(): void
```
simon-debruijn, line 45: And can the tests be in format `test_it_fails_when_integration_is_disabled`

app/Keycloak/Dto/Realm.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        return $this->internalName;
+    }
+
+    public function getPublicName(): string
+    {
+        return $this->publicName;
+    }
+
+    public static function getMasterRealm(): self
```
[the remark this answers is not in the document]
simon-debruijn: I do get why you call it `getMasterRealm` in this scenario, either is fine

simon-debruijn (approved):

app/Keycloak/Dto/Realm.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak\Dto;
```
LucWollants, line 5: I would even place this class and the `Config` class inside the `App\Keycloack` namespace. That was also done for all other domains likes Insightly, Auth0 and UiTiD.

app/Keycloak/TokenStrategy/ClientCredentials.php
```diff
@@ -0,0 +1,66 @@
...
+            try {
+                if (!$this->config->isEnabled()) {
```
LucWollants, line 32: I am not sure if this is the place to check the feature toggle if Keycloack is enabled or not. If you think about the final solution, it will be the creation of keys we want to enable. So looking at all other integrations this needs to be done inside the provider like here https://github.com/cultuurnet/publiq-platform/blob/main/app/Auth0/Auth0ServiceProvider.php#L92

app/Keycloak/TokenStrategy/ClientCredentials.php
LucWollants: I'm also not sure if this approach is good. What if you change the realm? Then you still get the token of the previous realm.

tests/Keycloak/TokenStrategy/ClientCredentialsTest.php
LucWollants: The name can also be more descriptive, something like `test_it_returns_a_valid_token`

app/Keycloak/TokenStrategy/ClientCredentials.php
```diff
@@ -0,0 +1,66 @@
...
+
+/*
+ * LIMITATION: This class currently does not refresh the token automatically.
+ * For normal usage this should be ok, but if we ever implement long-running CLI processes this will need to be improved.
+ * */
+final class ClientCredentials implements TokenStrategy
+{
+    private ?string $accessToken = null;
+
+    public function __construct(
```
LucWollants, line 22: This class can benefit from a logger once it is all plugged in.

app/Keycloak/Collection/RealmCollection.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace App\Keycloak\Collection;
+
+use App\Keycloak\Dto\Realm;
+use Ramsey\Collection\AbstractCollection;
+
+/**
+ * @extends AbstractCollection<Realm>
+ */
+final class RealmCollection extends AbstractCollection
```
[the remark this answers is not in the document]
grubolsch: Personally I find Ramsey collection better because it actually checks the type on runtime, and is not just a docblock.
But given that we are in Laravel I changed it.

app/Keycloak/TokenStrategy/ClientCredentials.php
grubolsch: Removed the check

app/Keycloak/RealmCollection.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace App\Keycloak\Collection;
+
+use App\Keycloak\Realm;
+use Illuminate\Support\Collection;
+
+/**
+ * @extends Collection<int, Realm>
+ */
+final class RealmCollection extends Collection
```
LucWollants, line 12: Another naming pattern can be to just use the plural form `Realms`, but no need to change it. (This plural form follows the approach of not suffixing classes with Interface or Exception)

PR #1064, opened 2024-05-03
PPF-437 Fetch clients keycloak

grubolsch:
This is just the start of this feature, more PR's will follow.

### Added
- Fetching (all) clients from keycloak

### Todo in a later PR
- Add UI to see these clients
- Add useful filter

---
Ticket: https://jira.uitdatabank.be/browse/PPF-437

### DIFF
--- a/app/Keycloak/ApiClient.php
+++ b/app/Keycloak/ApiClient.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak;
+
+use App\Json;
+use App\Keycloak\Collection\ClientCollection;
+use App\Keycloak\Dto\Client;
+use App\Keycloak\Dto\Config;
+use App\Keycloak\Dto\Realm;
+use App\Keycloak\Exception\KeyCloakApiFailed;
+use App\Keycloak\TokenStrategy\TokenStrategy;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\GuzzleException;
+use JsonException;
+use Psr\Log\LoggerInterface;
+
+final readonly class ApiClient
+{
+    public function __construct(
+        private ClientInterface $client,
+        private Config $config,
+        private TokenStrategy $tokenStrategy,
+        private LoggerInterface $logger
+    ) {
+    }
+
+    /**
+     * @return ClientCollection[]
+     */
+    public function fetchClients(string $clientId): array
+    {
+        $output = [];
+        foreach ($this->config->getRealms() as $realm) {
+            $output[$realm->getInternalName()] = $this->fetchClientsPerRealm($realm, $clientId);
+        }
+
+        return $output;
+    }
+
+    private function fetchClientsPerRealm(Realm $realm, string $clientId): ClientCollection
+    {
+        $clientCollection = new ClientCollection();
+
+        try {
+            $response = $this->client->request(
+                'GET',
+                $this->config->getBaseUrl() . 'admin/realms/' . $realm->getInternalName() . '/clients',
+                [
+                    // @todo Not sure if I need to filter on clientId, or some meta information field - more a proof of concept
+                    'query' => ['clientId' => $clientId],
+                    'headers' => $this->getHeaders(),
+                ]
+            );
+
+            if ($response->getStatusCode() !== 200) {
+                throw KeyCloakApiFailed::failedToFetchClient($realm, $response->getBody()->getContents());
+            }
+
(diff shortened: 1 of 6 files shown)


app/Keycloak/ApiClient.php
```diff
@@ -0,0 +1,78 @@
...
+                throw KeyCloakApiFailed::failedToFetchClient($realm, $response->getBody()->getContents());
+            }
+
+            $body = $response->getBody()->getContents();
+            foreach (Json::decodeAssociatively($body) as $client) {
+                $clientCollection->add(Client::fromJson($client));
+            }
+        } catch (KeyCloakApiFailed|JsonException|GuzzleException $e) {
+            $this->logger->error($e->getMessage());
+        }
```
simon-debruijn, line 67: I would just catch the Exception class here since there is no specific handling, just logging

app/Keycloak/Collection/ClientCollection.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak\Collection;
+
+use App\Keycloak\Dto\Client;
+use Ramsey\Collection\AbstractCollection;
```
simon-debruijn, line 8: Since we are in a laravel codebase, doesn't it make more sense to use `Illuminate\Support\Collection`?

app/Keycloak/Dto/Client.php
```diff
@@ -0,0 +1,244 @@
...
+    private bool $frontchannelLogout;
+    private string $protocol;
+    private array $attributes;
+    private array $authenticationFlowBindingOverrides;
+    private bool $fullScopeAllowed;
+    private int $nodeReRegistrationTimeout;
+    private array $protocolMappers;
+    private array $defaultClientScopes;
+    private array $optionalClientScopes;
+    private array $access;
```
simon-debruijn, line 40: Maybe these properties can be grouped more into other classes. e.g. scopes, urls, flags

tests/Keycloak/ApiClientTest.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->config = new Config(
+            true,
+            'https://keycloak.example.com/',
+            'php_client',
+            'a_true_secret',
+            new Realm('uitidpoc', 'Acceptance')
+        );
+    }
+
+    public function testHappyPath(): void
```
simon-debruijn, line 33: Test names are written in snake_cased in publiq-platform

app/Keycloak/Collection/ClientCollection.php
simon-debruijn: Same goes for all the Collections used

simon-debruijn (approved):

app/Keycloak/ApiClient.php
```diff
@@ -0,0 +1,78 @@
...
+        private Config $config,
+        private TokenStrategy $tokenStrategy,
+        private LoggerInterface $logger
+    ) {
+    }
+
+    /**
+     * @return ClientCollection[]
+     */
+    public function fetchClients(string $clientId): array
```
LucWollants, line 32: Is the clientId not part of the config?

app/Keycloak/ApiClient.php
```diff
@@ -0,0 +1,78 @@
...
+            $this->logger->error($e->getMessage());
+        }
+
+        return $clientCollection;
+    }
+
+    private function getHeaders(): array
+    {
+        return [
+            'Authorization' => 'Bearer ' . $this->tokenStrategy->fetchToken(Realm::getMasterRealm()),
```
LucWollants, line 75: I would expect this to get a token from the active realm from the loop here https://github.com/cultuurnet/publiq-platform/pull/1064/files#diff-713965e8ea8f9297cada7462bf7f5b78e2f413a8287ee95cdb8337291447c289R35

app/Keycloak/Dto/Client.php
LucWollants: I also wonder if we need all these values in the context of publiq-platform? If we take the scope of this feature into account the final solution needs to create clients inside all the configured Keycloack environments like we do for Auth0 and UiTiD. We don't need to know all the details, just the client id and the secret.

LucWollants: I know you work step by step, which is great, but I think we miss an essential part and that is how we will store the created clients inside publiq-platform. Because those credentials are stored internally.

grubolsch:
I am going to park this PR for now in draft, because we might not need this based on my discussion with Luc.

PR #1081, opened 2024-05-08
PPF-435 Added scope config for Keycloak

grubolsch:
### Added
Scope in Keycloak is wat determines what a client can do, in Publiq Platform this is dependant on the Integration type.
We need to know the scope id in Keycloak to add this to our client.

In a later PR an API call we be that adds this scope to a client.

Contains an extra helper for the unit tests - might seem overkill now but I will need to create a lot of integrations for later Unit tests. ;-)

---

### DIFF
--- a/app/Keycloak/ScopeConfig.php
+++ b/app/Keycloak/ScopeConfig.php
@@ -8,21 +8,21 @@ use App\Domain\Integrations\Integration;
 use App\Domain\Integrations\IntegrationType;
 use Ramsey\Uuid\UuidInterface;
 
-final class ScopeConfig
+readonly final class ScopeConfig
 {
     public function __construct(
-        public UuidInterface $searchApi,
-        public UuidInterface $entryApi,
-        public UuidInterface $widgets,
+        public UuidInterface $searchApiScopeId,
+        public UuidInterface $entryApiScopeId,
+        public UuidInterface $ScopeId,
     ) {
     }
 
     public function getScopeIdFromIntegrationType(Integration $integration): UuidInterface
     {
         return match ($integration->type) {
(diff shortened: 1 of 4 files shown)


app/Keycloak/ScopeConfig.php
```diff
@@ -0,0 +1,28 @@
...
+namespace App\Keycloak;
+
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\IntegrationType;
+use Ramsey\Uuid\UuidInterface;
+
+final class ScopeConfig
+{
+    public function __construct(
+        public UuidInterface $searchApi,
```
LucWollants, line 14: I was first confused about what this was, but it seems this is the scopeId of integration type SearchApi. Maybe make the name a bit more explicit?

tests/IntegrationHelper.php
```diff
@@ -0,0 +1,28 @@
...
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\IntegrationPartnerStatus;
+use App\Domain\Integrations\IntegrationStatus;
+use App\Domain\Integrations\IntegrationType;
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+
+trait IntegrationHelper
+{
+    public function createIntegration(UuidInterface $integrationId, array $options = []): Integration
```
LucWollants, line 16: This is often called `giventhereIsAnIntegration`, see https://github.com/cultuurnet/publiq-platform/blob/main/tests/Domain/Integrations/Controllers/IntegrationControllerTest.php#L951

LucWollants (approved):
I already approved because I only have smaller remarks.

PR #1083, opened 2024-05-08
PPF-435 Add scope to client

grubolsch:
### Added
Add scope (widgets / sapi3 / entry api) to Keycloak client

---

### DIFF
--- a/app/Keycloak/Service/ApiClient.php
+++ b/app/Keycloak/Service/ApiClient.php
@@ -10,6 +10,7 @@ use App\Keycloak\Client\KeycloakClientWithBearer;
 use App\Keycloak\Exception\KeyCloakApiFailed;
 use App\Keycloak\Realm;
 use Exception;
+use GuzzleHttp\Exception\GuzzleException;
 use GuzzleHttp\Psr7\Request;
 use Psr\Log\LoggerInterface;
 use Ramsey\Uuid\Uuid;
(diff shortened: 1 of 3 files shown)


app/Keycloak/Service/ApiClient.php
```diff
@@ -48,4 +49,22 @@ public function createClient(Realm $realm, Integration $integration): UuidInterf
 
         return $id;
     }
+
+    public function addScopeToClient(Realm $realm, UuidInterface $clientId, UuidInterface $scopeId): void
```
LucWollants, line 53: When creating the client in Keycloack it is not possible to set the correct scope?

grubolsch: I am afraid not, that is also not possible in the UI of keycloak, you must first create a client to add scopes.
You can set default scopes in the general config, but given that it varies which role we want, we cannot use this.

Here you can see the fields we can sent: https://www.keycloak.org/docs-api/latest/rest-api/index.html#ClientRepresentation

LucWollants: It's not that the fields `defaultClientScopes` or `optionalClientScopes` can hold the scopes?

grubolsch: I just tried and this does not work.
I also did some additional research and all google post say the same, you have to call the extra api endpoint after creating the client. This is also how it works in the GUI of keycloak.

LucWollants (approved):

PR #1084, opened 2024-05-08
PPF-435 Fetching client info from Keycloak

grubolsch:
### Added
Fetching client info from Keycloak.
This is needed because when creating a new client with the API you don't get any response, so you don't know the created secret.

Later this fetched info will be stored in our db.

---


LucWollants (approved):
Looks good!

PR #1085, opened 2024-05-08
PPF-435 Complete flow creating client keycloak

grubolsch:
### Added
This pr adds the bootstrapping / service provider of the ApiClient of Keycloak, for later use.

It also contains a class CreateClientFlow, which runs the entire flow of
- creating a client
- add the correct scope
- retrieving the client information (secret) for later storage

It also has a CLI command to create a client from an integration.
While this was not requested, it is useful as a proof of concept / handy during testing.

# Related
https://github.com/cultuurnet/appconfig/pull/616

---

### DIFF
--- a/app/Keycloak/Service/CreateClientFlow.php
+++ b/app/Keycloak/Service/CreateClientFlow.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak\Service;
+
+use App\Domain\Integrations\Integration;
+use App\Keycloak\ClientCollection;
+use App\Keycloak\Config;
+use App\Keycloak\Exception\KeyCloakApiFailed;
+use App\Keycloak\ScopeConfig;
+use Psr\Log\LoggerInterface;
+
+final readonly class CreateClientFlow
+{
+    public function __construct(
+        private ApiClient $client,
+        private Config $config,
+        private ScopeConfig $scopeConfig,
+        private LoggerInterface $logger
+    ) {
+    }
+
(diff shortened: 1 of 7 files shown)


grubolsch:
Looks like this needs actual config to work.

app/Keycloak/Service/CreateClientFlow.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function __construct(
+        private ApiClient $client,
+        private Config $config,
+        private ScopeConfig $scopeConfig,
+        private LoggerInterface $logger
+    ) {
+    }
+
+    public function createClientsForIntegration(Integration $integration): ClientCollection
```
LucWollants, line 24: Can this become part of the `ApiClient`? I don't think there is added value of having this extra class.

grubolsch: I would prefer not too because later on I plan to also add the logic to save the Eloquent model to the database for each client. 
I personally like to have several more smaller classes than 1 large class, but if you want me too I will merge them.

LucWollants: A dedicated class per action (create, update, activate, block) is fine for me, but I would search for a better name than `CreateClientFlow::createClientsForIntegration`. Maybe you can apply some naming conventions like a command handler `CreateClientHandler:handle`

grubolsch: great idea, will change the naming!

simon-debruijn (approved):

LucWollants (approved):

PR #1086, opened 2024-05-10
Add the email of helpdesk to admins

grubolsch:
### Added
Robin asked me to add the emails of the helpdesk to the admin login.


LucWollants:
@grubolsch do they need Nova access (admin backend) or Horizon access (queue monitor)? Because now you gave them Horizon access.

JonasVHG (changes requested):
They should be added to `config/nova.php` not `config/horizon.php`

PR #1087, opened 2024-05-10
PPF-435 Add elequant models

grubolsch:
### Added
Add eloquent / nova resources to show Keycloak integrations

This pr does not include the actual possibility to create keycloak integrations from the UI.

---


app/Keycloak/Repositories/KeycloakClientRepository.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak\Repositories;
+
+interface KeycloakClientRepository
```
LucWollants, line 7: This misses all the methods that need to be implemented inside the eloquent concrete implementation.

app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
```diff
@@ -0,0 +1,79 @@
...
+namespace App\Keycloak\Repositories;
+
+use App\Keycloak\Client;
+use App\Keycloak\Models\KeycloakClientModel;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Support\Facades\DB;
+use Ramsey\Uuid\UuidInterface;
+
+final class EloquentKeycloakClientRepository implements KeycloakClientRepository
```
LucWollants, line 14: This misses unit tests

grubolsch:
Processed all feedback, added extra tests.

app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
```diff
@@ -24,13 +25,12 @@ public function save(Client ...$clients): void
                 KeycloakClientModel::query()
                     ->updateOrCreate(
                         [
-                            'client_id' => $client->clientId,
+                            'client_secret' => $client->clientSecret, // secret could actually change because of secret rotation
```
LucWollants, line 28: Is this something we want? What happens to an existing integration, will it stop working because the secret is different?

LucWollants: I also think this code is wrong. The first array is used to find the entity and the second array is used to update the values. See https://laravel.com/docs/11.x/eloquent#updates

tests/Keycloak/Repositories/EloquentKeycloakClientRepositoryTest.php
```diff
@@ -0,0 +1,191 @@
...
+use App\Keycloak\Client;
+use App\Keycloak\Realm;
+use App\Keycloak\RealmCollection;
+use App\Keycloak\Repositories\EloquentKeycloakClientRepository;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Ramsey\Uuid\Uuid;
+use Tests\TestCase;
+
+final class EloquentKeycloakClientRepositoryTest extends TestCase
+{
```
LucWollants, line 16: A test for an update is missing

app/Keycloak/Service/ApiClientInterface.php
```diff
@@ -0,0 +1,19 @@
...
+declare(strict_types=1);
+
+namespace App\Keycloak\Service;
+
+use App\Domain\Integrations\Integration;
+use App\Keycloak\Client;
+use App\Keycloak\Realm;
+use Ramsey\Uuid\UuidInterface;
+
+interface ApiClientInterface
```
LucWollants, line 12: I am not a fan of using the `interface` suffix. I think we completely avoided this in this repository.

app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
grubolsch: Respectfully, I don't think so.

> 
> In the example below, if a flight exists with a departure location of Oakland and a destination location of San Diego, its price and discounted columns will be updated. If no such flight exists, a new flight will be created which has the attributes resulting from merging the first argument array with the second argument array:
> 
> $flight = Flight::updateOrCreate(
>     ['departure' => 'Oakland', 'destination' => 'San Diego'],
>     ['price' => 99, 'discounted' => 1]
> );
> 

If you look at the auth0 repo you can also see the same structure
```

Auth0ClientModel::query()
                    ->updateOrCreate(
                        [
                            'auth0_client_id' => $auth0Client->clientId,
                            'auth0_tenant' => $tenant,
                        ],
                        [
                            'id' => $auth0Client->id->toString(),
                            'integration_id' => $auth0Client->integrationId->toString(),
                            'auth0_client_id' => $auth0Client->clientId,
                            'auth0_client_secret' => $auth0Client->clientSecret,
                            'auth0_tenant' => $tenant,
                        ]
                    );
```

LucWollants:
```
'client_secret' => $client->clientSecret, // secret could actually change because of secret rotation
```

If the secret is different, this would mean that the client will not be found and a new one would be created.

tests/Keycloak/Repositories/EloquentKeycloakClientRepositoryTest.php
grubolsch: No longer needed because we don't do updates anymore

app/Nova/Resources/Integration.php
```diff
@@ -152,6 +152,7 @@ public function fields(NovaRequest $request): array
 
             HasMany::make('UiTiD v1 Consumer Credentials', 'uiTiDv1Consumers', UiTiDv1::class),
             HasMany::make('UiTiD v2 Client Credentials (Auth0)', 'auth0Clients', Auth0Client::class),
+            HasMany::make('Keycloak client Credentials', 'keycloakClients', KeycloakClient::class),
```
LucWollants, line 155: This is maybe something to hide behind the feature flag?

LucWollants (approved):

app/Nova/Resources/Integration.php
grubolsch: I will pick that up in a next pr

PR #1103, opened 2024-05-15
PPF-436 Create listeners

grubolsch:
### Added
This PR adds the listener when an integration is created, to automatically create the integration.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-436


app/Keycloak/Client.php
```diff
@@ -34,6 +34,6 @@ public static function createFromJson(Realm $realm, UuidInterface $integrationId
 
     public function getKeycloakUrl(string $baseUrl): string
     {
-        return sprintf('%s/admin/master/console/#/%s/clients/%s/settings', $baseUrl, $this->realm->internalName, $this->integrationId->toString());
+        return $baseUrl . sprintf('admin/master/console/#/%s/clients/%s/settings', $this->realm->internalName, $this->id->toString());
```
LucWollants, line 37: Is it concatenation or interpolation? Mixing them feels a bit weird.

app/Keycloak/KeycloakServiceProvider.php
```diff
@@ -48,5 +54,20 @@ public function register(): void
...
         });
+
+        $this->app->singleton(KeycloakClientRepository::class, function () {
+            return $this->app->get(EloquentKeycloakClientRepository::class);
+        });
+
+        $this->applyEvents();
+    }
+
+    private function applyEvents(): void
```
LucWollants, line 65: Maybe something like `bootstrapEventHandling` is better?

app/Keycloak/KeycloakServiceProvider.php
```diff
@@ -48,5 +54,20 @@ public function register(): void
...
+    {
+        if (!$this->app->get(Config::class)->isEnabled) {
```
LucWollants, line 67: Nit: It is also possible to create the Config class on the stack on line 40 and then pass it in here. But both are fine.

app/Keycloak/Service/CreateClientHandler.php
LucWollants, line 1: This seems to be an empty file?

config/keycloak.php
```diff
@@ -3,7 +3,7 @@
 declare(strict_types=1);
 
 return [
-    'enabled' => env('KEYCLOAK_ENABLED', false),
+    'enabled' => env('KEYCLOAK_ENABLE', false),
```
LucWollants, line 6: I guess the old name was better?

LucWollants (approved):
I already approved, just some minor remarks

config/keycloak.php
grubolsch: Fine, but that does request an extra pr: https://github.com/cultuurnet/appconfig/pull/632

PR #1119, opened 2024-05-21
PPF-439 status api keycloak

grubolsch:
### Added
Added the ability to enable/disable a client.
Also fetches the correct status in Keycloak to show to the user.

Screenshot 2024-05-21 at 11 22 05

---
Ticket: https://jira.uitdatabank.be/browse/PPF-439


app/Keycloak/Events/ClientDisabled.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak\Events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use Ramsey\Uuid\UuidInterface;
+
+final class ClientDisabled
```
LucWollants, line 10: Maybe we reuse the terminology from `Auth0` being `ClientBlocked` and `UiTiD` being `ConsumerBlocked` instead of using Enable/Disable because we are inside the publiq-platform domain space.

LucWollants (approved):
Looks good! Only I doubt a bit about using enable/disable instead of block/activate.

Maybe someone else has another opinion about this.

PR #1122, opened 2024-05-21
PPF-438 update existing client

grubolsch:
### Added
- When updating integration, we also update the matching keycloak clients.
- This means we also remove all scopes managed by publiq platform, to re-add them to the integration, to account for changes in the type.
- Also added some extra tests in KeycloakApiClientTest

---
Ticket: https://jira.uitdatabank.be/browse/PPF-438


app/Keycloak/Service/KeycloakApiClient.php
```diff
@@ -70,6 +77,31 @@ public function addScopeToClient(Realm $realm, UuidInterface $clientId, UuidInte
         }
     }
 
+
+    public function resetScopes(Client $client): void
```
LucWollants, line 81: Is this rather a `deleteScopes` instead of a `resetScopes`?

LucWollants (approved):
I already approved, but I left a comment on https://jira.publiq.be/browse/PPF-438
We need to rethink the changes required to change the type of an existing integration.

PR #1124, opened 2024-05-22
PPF-486 Sync integration uri

grubolsch:
### Added
Make sure when integration urls are created / changed that they are synced in Keycloak.

Screenshot 2024-05-22 at 10 39 35

---
Ticket: https://jira.uitdatabank.be/browse/PPF-486


app/Domain/Integrations/Events/IntegrationUpdated.php
```diff
@@ -14,4 +14,9 @@ final class IntegrationUpdated
     public function __construct(public readonly UuidInterface $id)
     {
     }
+
+    public function getIntegrationId(): UuidInterface
```
LucWollants, line 18: This is no longer used

app/Keycloak/KeycloakServiceProvider.php
```diff
@@ -84,6 +87,10 @@ private function bootstrapEventHandling(): void
 
         Event::listen(IntegrationCreated::class, [CreateClients::class, 'handle']);
         Event::listen(IntegrationUpdated::class, [UpdateClients::class, 'handle']);
-        Event::listen(IntegrationBlocked::class, [DisableClients::class, 'handle']);
+        Event::listen(IntegrationBlocked::class, [BlockClients::class, 'handle']);
```
LucWollants, line 90: This seems to include the wrong namespace?

app/Keycloak/Listeners/EnableClients.php
```diff
@@ -0,0 +1,56 @@
...
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Keycloak\Exception\KeyCloakApiFailed;
+use App\Keycloak\Repositories\KeycloakClientRepository;
+use App\Keycloak\Service\ApiClient;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+final class EnableClients implements ShouldQueue
```
LucWollants, line 17: I think you missed the linking of this new listener? This should have been done in another pull request?

app/Domain/Integrations/Environment.php
```diff
@@ -9,4 +9,5 @@ enum Environment: string
     case Acceptance = 'acc';
     case Testing = 'test';
     case Production = 'prod';
+    case Unknown = 'unknown';
```
LucWollants, line 12: Why do we need this extra environment?

app/Keycloak/Listeners/EnableClients.php
grubolsch: you are right, there currently is no action to re-enable all clients.
This regarding the ticket we discussed this morning.
I will park the enable code in a seperate draft pr

LucWollants: If this is a new class it also lacks some tests.

app/Keycloak/Service/IntegrationUrlConverter.php
```diff
@@ -0,0 +1,65 @@
...
+            // Only first parties can have redirect uri configured.
+            return $urls;
+        }
+
+        $urls = self::getCallbackUrl($integration, $client, $urls);
+        $urls = self::getLoginUrls($integration, $client, $urls);
+        return self::getLogoutUrls($integration, $client, $urls);
+    }
+
+    private static function getCallbackUrl(Integration $integration, Client $client, array $urls): array
```
LucWollants, line 38: These methods do more then get, they do also a set

app/Keycloak/Service/IntegrationUrlConverter.php
```diff
@@ -0,0 +1,65 @@
...
+        $urls = [
+            'baseUrl' => '',
+            'redirectUris' => [],
+            'attributes' => ['post.logout.redirect.uris' => ''],
+            'webOrigins' => [
+                '+', // This permits all origins of Valid Redirect URIs for CORS checks
+            ],
+        ];
+
+        if ($integration->partnerStatus !== IntegrationPartnerStatus::FIRST_PARTY) {
```
LucWollants, line 28: I wonder if we should do the same for Auth0 and UiTiD?

grubolsch: Discussed it briefly with Jonas here. UiTiDv1 does not have the concept of 1st vs 3th party clients, so no issues there according to Jonas. In auth0 it would be a nice check, but it does not give issues currently and I would not put time in it because it will be replaced in less then a year anyway.

app/Domain/Integrations/Environment.php
grubolsch: Removed it. The only downside is that the Master realm is now attached to the production realm, which is not entirely correct.
But the only place we use the Environment is to match the correct IntegrationUri to the correct realm, so it is not a problem.

LucWollants (approved):

PR #1129, opened 2024-05-22
PPF-488 Always send the groups when updating uitidv1

grubolsch:
### Fixed
Activating an integration removes the consumer groups in UiTID.
To fix this, I now always sent the groups when updating an UiTiDv1 consumer.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-488

### DIFF
--- a/tests/IntegrationHelper.php
+++ b/tests/IntegrationHelper.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests;
+
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\IntegrationPartnerStatus;
+use App\Domain\Integrations\IntegrationStatus;
+use App\Domain\Integrations\IntegrationType;
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+
+trait IntegrationHelper
+{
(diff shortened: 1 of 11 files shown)


tests/IntegrationHelper.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Tests;
+
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\IntegrationPartnerStatus;
+use App\Domain\Integrations\IntegrationStatus;
+use App\Domain\Integrations\IntegrationType;
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+
+trait IntegrationHelper
```
LucWollants, line 14: Maybe give this a more specific name? Something like `createIntegrationMock` like we have `CreatesMockUiTiDv1ClusterSDK`?

LucWollants (approved):
Very nice work!

tests/IntegrationHelper.php
grubolsch: Is "createIntegration" good instead of "createIntegrationMock" regarding your comment? I find including mock in the name wrong, because that is the opposite it does. It does not provide a mock, it provides a concrete implementation.

PR #1136, opened 2024-05-23
PPF-487 Create missing keycloak integrations

grubolsch:
### Added
Add an option to create missing keycloak integrations

---
Ticket: https://jira.uitdatabank.be/browse/PPF-487


app/Keycloak/Jobs/MissingClientsDetected.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Keycloak\Jobs;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use Ramsey\Uuid\UuidInterface;
+
+final class MissingClientsDetected
```
LucWollants, line 10: Based on the past tense of the verb `detect` this is rather an Event instead of a Job. So I would place this inside the `Events` folder.

It is a bit different from the `Auth0` namespace where `CreateMissingClients` is used and that is a `Job` with a dedicated handler `CreateMissingClientsHandler`. Here we have an event `MissingClientsDetected` and the listener `CreateMissingClients`

app/Keycloak/Listeners/CreateMissingClients.php
```diff
@@ -0,0 +1,60 @@
...
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Keycloak\Jobs\MissingClientsDetected;
+use App\Keycloak\Repositories\KeycloakClientRepository;
+use App\Keycloak\Service\CreateClientForRealms;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+final class CreateMissingClients implements ShouldQueue
```
LucWollants, line 16: I would bundle the `handle` methods from `CreateClients` and `CreateMissingClients`. As far as I can see the only difference is that `getMissingRealmsByIntegrationId` gets called.

So you can create two public methods `handleCreateClients` and `handleCreatingMissingClients` which both call a private method `handle` that does the common part.

app/Keycloak/Service/ApiClient.php
```diff
@@ -11,7 +11,7 @@
 
 interface ApiClient
```
LucWollants, line 12: Should this interface move to the `Client` namespace instead of sitting inside the `Service` namespace?

LucWollants: Also the concrete implementation `KeycloakApiClient` can move.

tests/Keycloak/Listeners/CreateClientsTest.php
```diff
@@ -8,30 +8,34 @@
...
 use PHPUnit\Framework\TestCase;
 use Psr\Log\LoggerInterface;
 use Ramsey\Uuid\Uuid;
 use Tests\IntegrationHelper;
-use Tests\Keycloak\KeycloakHelper;
 
 final class CreateClientsTest extends TestCase
 {
     use IntegrationHelper;
-    use KeycloakHelper;
```
LucWollants, line 28: Was this not something that got changed to something more specific than `Helper`?

tests/Keycloak/Listeners/CreateClientsTest.php
```diff
@@ -8,30 +8,34 @@
...
 use App\Keycloak\Service\ApiClient;
+use App\Keycloak\Service\CreateClientForRealms;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Psr\Log\LoggerInterface;
 use Ramsey\Uuid\Uuid;
 use Tests\IntegrationHelper;
-use Tests\Keycloak\KeycloakHelper;
 
 final class CreateClientsTest extends TestCase
```
LucWollants, line 25: I wonder, should this not extend from `Tests\TestCase`

tests/Keycloak/Listeners/CreateMissingClientsHandlerTest.php
```diff
@@ -0,0 +1,164 @@
...
+use App\Keycloak\ScopeConfig;
+use App\Keycloak\Service\ApiClient;
+use App\Keycloak\Service\CreateClientForRealms;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+use Ramsey\Uuid\Uuid;
+use Tests\IntegrationHelper;
+
+final class CreateMissingClientsHandlerTest extends TestCase
```
LucWollants, line 25: The name should be `CreateMissingClientsTest`

grubolsch: Test is removed because merged with CreateClientsTest

grubolsch:
Feedback processed

tests/Keycloak/CreatedKeycloakHttpClient.php
```diff
@@ -13,7 +13,7 @@
 use GuzzleHttp\RequestOptions;
 use Psr\Log\LoggerInterface;
 
-trait KeycloakHelper
+trait CreatedKeycloakHttpClient
```
LucWollants, line 16: Maybe the name should include `Factory`? I also notice that `$mock` is always provided?

LucWollants (approved):

PR #1137, opened 2024-05-24
PPF-477-Add polling to check for key visibility upgrade

vhande:
### Added
- polling to check for key visibility upgrade

---
Ticket: https://jira.uitdatabank.be/browse/PPF-477


resources/ts/types/Integration.ts
```diff
@@ -32,4 +37,5 @@ export type Integration = {
   authClients: AuthClient[];
   legacyAuthConsumers: LegacyAuthConsumer[];
   keyVisibility: KeyVisibility;
+  keyVisibilityUpgrade: KeyVisibilityUpgrade;
```
simon-debruijn, line 40:
```suggestion
  keyVisibilityUpgrade: KeyVisibilityUpgrade | null;
```

This can also be null no?

simon-debruijn (changes requested):

resources/ts/types/Integration.ts
LucWollants: This is indeed not applicable to all integrations, only the integrations for which v2 keys are requested.

resources/ts/Components/Integrations/Detail/CredentialsAuthClients.tsx
```diff
@@ -63,26 +70,31 @@ export const CredentialsAuthClients = ({
...
-              i18nKey="details.credentials.uitid_alert"
-              components={[
-                <Link
-                  key={t("details.credentials.uitid_alert")}
-                  href={t("details.credentials.uitid_alert_link")}
-                  className="text-publiq-blue-dark hover:underline mb-3"
-                />,
-              ]}
-            />
+        keyVisibility === KeyVisibility.v1 && !!keyVisibilityUpgrade ? (
```
Anahkiasen, line 73: The first part of the condition is already covered by the ternary but more generally I would avoid the nested ternary as I feel it's hard to parse mentally. Could we extract the subcondition to a variable perhaps?

resources/ts/Components/Integrations/Detail/Credentials.tsx
```diff
@@ -40,7 +41,11 @@ export const Credentials = ({
   const hasAnyCredentials = Boolean(
     legacyAuthConsumers.length || authClients.length
   );
-  usePolling(!hasAnyCredentials, { only: ["integration"] });
+  usePolling(
+    !hasAnyCredentials ||
+      (keyVisibility === KeyVisibility.v1 && !!keyVisibilityUpgrade),
```
Anahkiasen, line 46: Minor but I would maybe but the second part of the condition in a variable like `hasAnyCredentials` was for readability purposes

Anahkiasen (changes requested):
Mostly code readability comments but functionality wise all good 👍

simon-debruijn (approved):

resources/ts/Components/Integrations/Detail/CredentialsAuthClients.tsx
```diff
@@ -62,27 +70,32 @@ export const CredentialsAuthClients = ({
...
-              components={[
-                <Link
-                  key={t("details.credentials.uitid_alert")}
-                  href={t("details.credentials.uitid_alert_link")}
-                  className="text-publiq-blue-dark hover:underline mb-3"
-                />,
-              ]}
-            />
+      {isKeyVisibilityV1 ? (
+        isKeyVisibilityV1 && !!keyVisibilityUpgrade ? (
```
Anahkiasen, line 74: My previous comment still stands here since we check twice for `isKeyVisibilityV1`, since it cannot check the second condition without checking the first. But since it's a plain boolean check it's not costly at runtime so not really major, so good for me either way

Anahkiasen (approved):

PR #1154, opened 2024-05-27
PPF-500 Rename disable to block

grubolsch:
This change is based on a meeting and feedback from Corneel.

### Changed
- In the context of Keycloak, change every reference from Disable to block
- In the context of Keycloak, change every reference from Enable to unblock / Active

---
Ticket: https://jira.uitdatabank.be/browse/PPF-500

### DIFF
--- a/app/Keycloak/CachedKeycloakClientStatus.php
+++ b/app/Keycloak/CachedKeycloakClientStatus.php
@@ -22,7 +22,7 @@ final class CachedKeycloakClientStatus
         if(! isset($this->statuses[$uuid])) {
             $this->statuses[$uuid] = $this->apiClient->fetchIsClientEnabled($client->realm, $client->integrationId);
         } else {
-            $this->logger->info(self::class . '  - ' . $uuid . ': cache hit: ' . ($this->statuses[$uuid] ? 'Enabled' : 'Disabled'));
+            $this->logger->info(self::class . '  - ' . $uuid . ': cache hit: ' . ($this->statuses[$uuid] ? 'Active' : 'Blocked'));
         }
 
         return $this->statuses[$uuid];
--- a/app/Keycloak/CachedKeycloakClientStatus.php
+++ b/app/Keycloak/CachedKeycloakClientStatus.php
@@ -15,16 +15,16 @@ final class CachedKeycloakClientStatus
     {
     }
 
-    public function isClientEnabled(Client $client): bool
+    public function isClientBlocked(Client $client): bool
     {
         $uuid = $client->id->toString();
 
         if(! isset($this->statuses[$uuid])) {
-            $this->statuses[$uuid] = $this->apiClient->fetchIsClientEnabled($client->realm, $client->integrationId);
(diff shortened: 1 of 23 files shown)


JonasVHG (approved):

app/Keycloak/Events/ClientBlocked.php
```diff
@@ -7,7 +7,7 @@
 use Illuminate\Foundation\Events\Dispatchable;
 use Ramsey\Uuid\UuidInterface;
 
-final class ClientEnabled
+final class ClientBlocked
```
LucWollants, line 10: I think `ClientEnabled` needs to be `ClientUnblocked`?

LucWollants (changes requested):
I think some changes are done wrong. Normally every `Enable` should match an `Unblock` and every `Disable` should match a `Block`

app/Nova/Actions/Keycloak/BlockKeycloakClient.php
```diff
@@ -14,14 +14,14 @@
...
 use Laravel\Nova\Fields\ActionFields;
 
-final class EnableKeycloakClient extends Action
+final class BlockKeycloakClient extends Action
 {
     use InteractsWithQueue;
     use Queueable;
 
-    public $name = 'Enable Keycloak client';
+    public $name = 'Disable Keycloak client';
```
LucWollants, line 22: Here it got changed to Disable, that should be Block or Unblock

app/Nova/Actions/Keycloak/UnblockKeycloakClient.php
```diff
@@ -14,14 +14,14 @@
...
 use Laravel\Nova\Fields\ActionFields;
 
-final class DisableKeycloakClient extends Action
+final class UnblockKeycloakClient extends Action
 {
     use InteractsWithQueue;
     use Queueable;
 
-    public $name = 'Disable Keycloak client';
+    public $name = 'Enable Keycloak client';
```
LucWollants, line 22: Here it got changed to Enable, that should be Unblock

app/Keycloak/Client/ApiClient.php
```diff
@@ -19,9 +19,9 @@ public function fetchClient(Realm $realm, Integration $integration): Client;
 
     public function fetchIsClientEnabled(Realm $realm, UuidInterface $integrationId): bool;
```
LucWollants, line 20: Here is also still a reference to enabled.

app/Keycloak/CachedKeycloakClientStatus.php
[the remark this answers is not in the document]
grubolsch: Yeah, I agree. I did not rename this method with the thinking that this is getting close of the internals of Keycloak where it is related to the "enabled" flag.
I also wanted to prevent using a double negative.
I could call it "isClientActive"? would that work for you?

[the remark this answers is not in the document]
LucWollants: Maybe use `isClientBlocked`

PR #1159, opened 2024-05-28
PPF-493 Add client id to keycloakclient

grubolsch:
### Changed
In preparation of the bigger work of this ticket, I added a separate client_id and integration_id for all the keycloak clients.
This will allow us to have more than 1 client per integration on a single realm.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-493

### DIFF
--- a/app/Keycloak/Client/KeycloakApiClient.php
+++ b/app/Keycloak/Client/KeycloakApiClient.php
@@ -131,8 +131,6 @@ final readonly class KeycloakApiClient implements ApiClient
             $data = Json::decodeAssociatively($body);
             return Client::createFromJson($realm, $integration->id, $clientId, $data[0]);
         } catch (Throwable $e) {
-            Log::error($e->getLine() . '/' . $e->getMessage());
-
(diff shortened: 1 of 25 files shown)


simon-debruijn (approved):

app/Keycloak/Client/KeycloakApiClient.php
```diff
@@ -111,13 +112,13 @@ public function deleteScopes(Client $client): void
...
      */
-    public function fetchClient(Realm $realm, Integration $integration): Client
+    public function fetchClient(Realm $realm, Integration $integration, UuidInterface $clientId): Client
     {
         try {
             $response = $this->client->sendWithBearer(
                 new Request(
                     'GET',
-                    'admin/realms/' . $realm->internalName . '/clients?' . http_build_query(['clientId' => $integration->id->toString()])
+                    'admin/realms/' . $realm->internalName . '/clients?' . http_build_query(['clientId' => $clientId->toString()])
```
LucWollants, line 121: According the API docs all calls can be done with the id field instead of the clientId

PR #1161, opened 2024-05-30
Ppf 485 unblock integration migration

JonasVHG:
### Added

- `2024_05_28_105201_integrations_previous_statuses`: Migration for new database table to store `previousStatus`
- `IntegrationPreviousStatusModel`: Model for new Database Table

### Changed

- `IntegrationModel`: save `PreviousStatus` during `block()`

### Fixed

- The previous status is saved during a `block`

### TODO in following PR's:

- `unblock()`
- unblock on `projectaanvraag`, `Keycloak`, `UiTID` & `Auth0`

---
Ticket: https://jira.publiq.be/browse/PPF-485


database/migrations/2024_05_28_105201_integrations_previous_statuses.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('integrations_previous_statuses', function (Blueprint $table) {
+            $table->uuid('id')->primary();
```
LucWollants, line 13: I don't think we need both an `id` and an `integration_id` column. The `id` column can just be used to store the integration ID and that makes it also unique.

LucWollants (approved):

PR #1162, opened 2024-05-30
PPF-493 Make realm contain the api uri

grubolsch:
### Context
Currently we assumed 1 keycloak instance running multiple realms (acc, test, prod).

After a conversation with Erwin this was confirmed as false - each environment will have it's own instance. This will contain 2 realms, a master realm and a realm for acc OR test OR prod.

#### Consequences:

The URI has to be moved from the Config to the Realm object
The ScopeConfig should be dependent on the chosen Realm
When working with the master realm for requesting the token, the correct environment has to be given.

### Changed
+ The URI and secret was moved from the Config to the Realm object
+ When working with the master realm for requesting the token, the correct environment matching the original realm is given. (for example a request from Test should contact the Master on test server).
+ MINOR - Made names of environments (acc, test, prod) consistent with the other clients (uitid, auth0).
+ MINOR - Always use id instead of client id for API calls.

### Not in this PR
The ScopeConfig should be dependent on the chosen Realm
ClientId should be a string not a UUID because of migration from auth0.

### Config changes PR
https://github.com/cultuurnet/appconfig/pull/645

---
Ticket: https://jira.uitdatabank.be/browse/PPF-493


app/Domain/Integrations/Models/IntegrationModel.php
```diff
@@ -266,7 +267,9 @@ public function hasMissingUiTiDv1Consumers(): bool
 
     public function hasMissingKeycloakConsumers(): bool
     {
-        return $this->keycloakClients()->count() < count(RealmCollection::getRealms());
+        /** @var Config $config */
+        $config = App::get(Config::class);
```
LucWollants, line 271: You could directly use `Environment` here as was also done for UiTiD and Auth0

app/Keycloak/Config.php
```diff
@@ -6,24 +6,9 @@
 
 final readonly class Config
```
LucWollants, line 7: I no longer see a need for this class. I would directly access the feature flag like we do for Auth0 and UiTiD and in other places use the handy RealmCollection.

app/Keycloak/Models/KeycloakClientModel.php
```diff
@@ -28,13 +29,15 @@ final class KeycloakClientModel extends UuidModel
...
+        $config = App::get(Config::class);
+
         return new Client(
-            // Trying to use magic getters for eg $this->id gives a 0 back
             Uuid::fromString($this->id),
             Uuid::fromString($this->integration_id),
             Uuid::fromString($this->client_id),
             $this->client_secret,
-            RealmCollection::fromPublicName($this->realm)
+            $config->realms->fromPublicName($this->realm)
```
LucWollants, line 40: I think the solutions from `Auth0Client` and `UiTiDv1Consumer` are more elegant. They use the enum `Auth0Tenant` and `UiTiDv1Environment` instead of a full domain object. When taking that approach you also have the added that you will not expose the key and secret of a realm to the UI

app/Nova/Resources/KeycloakClient.php
```diff
@@ -63,7 +62,7 @@ public function fields(NovaRequest $request): array
             Select::make('realm')
                 ->readonly()
                 ->filterable()
-                ->options(RealmCollection::asArray()),
+                ->options($this->getConfig()->realms->asArray()),
```
LucWollants, line 65: Again, look at the implementation for Auth0 and UiTiD

config/keycloak.php
```diff
@@ -4,9 +4,26 @@
 
 return [
     'enabled' => env('KEYCLOAK_ENABLED', false),
-    'base_url' => env('KEYCLOAK_BASE_URL', ''),
-    'client_id' => env('KEYCLOAK_CLIENT_ID', ''),
-    'client_secret' => env('KEYCLOAK_CLIENT_SECRET', ''),
+    'environments' => [
+        'acc' => [
```
LucWollants, line 8: Here you use `acc` as the array key, but `STAG` in the name of the env variable

app/Keycloak/Models/KeycloakClientModel.php
grubolsch: I do not really see how I am exposing the key and secret to the ui, but I might not be aware of an Nova feature. Where could you see the secret of a realm in the ui?

I do not find this an an un-elegant design, the Realm is used a lot in other classes, for example the KeycloakApiClient to know to which realm to sent requests. 
If I remove it here I would have to add code in all these places to transform the Enum to a valid Realm object.

LucWollants: In the context of the `KeycloakApiClient` you indeed need the `Realm` because you need to know the URL and the credentials. But in the context of a saved `KeycloackClient you don't need that extra information, you only need to know the environment.

This is also a hot topic in the context of DDD to pick meaningful boundaries or bounded contexts:
- make the boundary too small and you will miss information, so you need a lot of other domain objects to get enough information
- make the boundary too big and you will be dragging along unimportant information and have overhead in creating your domain objects

app/Keycloak/KeycloakServiceProvider.php
```diff
@@ -36,27 +36,15 @@ public function register(): void
...
-                        $this->app->get(LoggerInterface::class)
+                        $this->app->get(LoggerInterface::class),
                     )
                 ),
                 $this->app->get(ScopeConfig::class),
                 $this->app->get(LoggerInterface::class),
             );
         });
 
-        $this->app->singleton(Config::class, function () {
```
LucWollants, line 50: You could make the `RealmCollection` a singleton and inject it where needed.

app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
```diff
@@ -78,7 +79,9 @@ public function getMissingRealmsByIntegrationId(UuidInterface $integrationId): R
     {
```
LucWollants, line 79: If you make this return the missing environments you make the code much easier. Because inside the `handleCreatingMissingClients` you have the list of realms and you just search for the missing inside the already constructed list.

app/Models/EnvironmentCollection.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace App\Models;
+
+use App\Domain\Integrations\Environment;
+use Illuminate\Support\Collection;
+
+/**
+ * @extends Collection<int, Environment>
+ */
+final class EnvironmentCollection extends Collection
```
LucWollants, line 13: This is not a model.

app/Keycloak/Client/KeycloakApiClient.php
```diff
@@ -31,18 +30,17 @@ public function __construct(
     /**
      * @throws KeyCloakApiFailed
      */
-    public function createClient(Realm $realm, Integration $integration, UuidInterface $clientId): void
+    public function createClient(Realm $realm, Integration $integration, UuidInterface $id): void
```
LucWollants, line 33: The `ApiClient` interface still has `clientId`

app/Keycloak/Client/KeycloakApiClient.php
```diff
@@ -63,20 +61,21 @@ public function createClient(Realm $realm, Integration $integration, UuidInterfa
...
             $response = $this->client->sendWithBearer(
                 new Request(
                     'PUT',
-                    sprintf('admin/realms/%s/clients/%s/default-client-scopes/%s', $realm->internalName, $clientId->toString(), $scopeId->toString())
-                )
+                    sprintf('admin/realms/%s/clients/%s/default-client-scopes/%s', $client->getRealm()->internalName, $client->id->toString(), $scopeId->toString())
+                ),
+                $client->getRealm()
             );
         } catch (GuzzleException $e) {
```
LucWollants, line 80: A special reason why this catches a `GuzzleException` and not a `Throwable` like all other methods?

LucWollants (approved):

PR #1165, opened 2024-05-30
PPF-485 unblock insightly

JonasVHG:
### Added
-  `UnblockOpportunity`
- `UnblockProject`
- `IntegrationStatusConverter`

### Changed

- `IntegrationModel`: dispatch `IntegrationUnblocked`
- `InsightlyServiceProvider`: add handle for `UnblockProject` & `UnblockOpportunity`

### Fixed

- `Makefile`: fix `Insightly`-test
-  `Integrations` can now be unblocked in `Insightly`

---
Ticket: https://jira.publiq.be/browse/PPF-485


LucWollants (changes requested):
I would use the previous state of the integration to determine which state and stage to set. See this table for correct mapping https://confluence.publiq.be/pages/viewpage.action?pageId=70686627#publiqplatformterminology-Status

JonasVHG:
> I would use the previous state of the integration to determine which state and stage to set. See this table for correct mapping https://confluence.publiq.be/pages/viewpage.action?pageId=70686627#publiqplatformterminology-Status

There were some errors in that table, also the difference between `ProjectStage` & `ProjectState` isn't really clear.
But unless it's a bug, The `ProjectStage` remains `Live` after a block.

LucWollants:
> > I would use the previous state of the integration to determine which state and stage to set. See this table for correct mapping https://confluence.publiq.be/pages/viewpage.action?pageId=70686627#publiqplatformterminology-Status
> 
> There were some errors in that table, also the difference between `ProjectStage` & `ProjectState` isn't really clear. But unless it's a bug, The `ProjectStage` remains `Live` after a block.

There is some extra information about stage and state on the same Confluence page: https://confluence.publiq.be/pages/viewpage.action?pageId=70686627#publiqplatformterminology-Pipelinestage

app/Insightly/IntegrationStatusConverter.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        return match ($status) {
+            IntegrationStatus::Draft, IntegrationStatus::PendingApprovalIntegration, IntegrationStatus::PendingApprovalPayment => OpportunityState::OPEN,
+            IntegrationStatus::Active => OpportunityState::WON,
+            IntegrationStatus::Blocked => OpportunityState::SUSPENDED,
+            default => null,
+        };
+    }
+
+    public static function getProjectStage(IntegrationStatus $status): ?ProjectStage
```
LucWollants, line 24: I would remove this method

app/Insightly/IntegrationStatusConverter.php
```diff
@@ -0,0 +1,40 @@
...
+namespace App\Insightly;
+
+use App\Domain\Integrations\IntegrationStatus;
+use App\Insightly\Objects\OpportunityState;
+use App\Insightly\Objects\ProjectStage;
+use App\Insightly\Objects\ProjectState;
+
+final class IntegrationStatusConverter
+{
+    public static function getOpportunitySate(IntegrationStatus $status): ?OpportunityState
```
LucWollants, line 14:
```suggestion
    public static function getOpportunityState(IntegrationStatus $status): ?OpportunityState
```

tests/Insightly/Listeners/UnblockOpportunityTest.php
```diff
@@ -0,0 +1,85 @@
...
+    protected function setUp(): void
+    {
+        $this->insightlyMappingRepository = $this->createMock(InsightlyMappingRepository::class);
+
+        $this->integrationRepository = $this->createMock(IntegrationRepository::class);
+
+        $this->mockCrmClient();
+
+        $this->unblockOpportunity = new UnblockOpportunity(
+            $this->insightlyClient,
```
LucWollants, line 43: I seem to miss the asserts/expects for the mocked Insightly client?

JonasVHG: Isn't that handled by `MockInsightlyClient` in https://github.com/cultuurnet/publiq-platform/blob/1541440b38978063079684b9761d3eca202d464f/tests/MockInsightlyClient.php#L38?

 There can only be an `opportunityResource`, which has an `expects()` if the `InsightlyClient` was called

LucWollants: There is no `expects` method added, only a `method` and `willReturn`.
See `ActivateProjectTest` for an example how the mocked resources are then used with an `expects`

LucWollants: Please discard this remark, the `expects` on the `opportunityResource` itself is enough. There is indeed no added value in calling the `expects` on `insightlyClient`.

LucWollants (approved):
Nice and clean!

PR #1194, opened 2024-06-11
Fixed UiTPAS Free plan not being selected by default

Anahkiasen:
### Fixed
- Fixed UiTPAS Free plan not being selected by default

---

Ticket: https://jira.publiq.be/browse/PPF-498


vhande (approved):

resources/ts/Pages/Integrations/New.tsx
```diff
@@ -77,6 +67,17 @@ const New = ({ subscriptions }: Props) => {
...
+  useEffect(() => {
+    const freeSubscriptionId = subscriptions.find(
+      (subscription) =>
+        subscription.integrationType === activeType &&
+        subscription.category === SubscriptionCategory.Free
+    )?.id;
+
+    setData("subscriptionId", freeSubscriptionId ?? "");
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [activeType, subscriptions]);
```
simon-debruijn, line 79: Is setData the missing dependency?

simon-debruijn (approved):

resources/ts/Pages/Integrations/New.tsx
Anahkiasen: Yeah it's the same issue as for all the form related methods they seem to completely mess with useEffect, but I wasn't sure if it was worth adding the same comment about it as in the other files :/ Im not sure if we're doing something wrong with useForm or something

PR #1198, opened 2024-06-12
PPF-508 dd migration auth0 -> keycloak clients

grubolsch:
### Added
Added a console script that migrates all auth0 clients that don't have a match in the keycloak table to Keycloak.

This PR is only about moving the auth0 clients in the database to keycloak clients. 
**It does not create these clients in Keycloak.**

It also has an option to start importing all clients from a certain date - in case we will do the migration in different milestones.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-508

### DIFF
--- a/app/Console/Commands/Migrations/MigrateAuth0ToKeycloakClients.php
+++ b/app/Console/Commands/Migrations/MigrateAuth0ToKeycloakClients.php
@@ -1,0 +1,88 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Console\Commands\Migrations;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+
+final class MigrateAuth0ToKeycloakClients extends Command
+{
+    protected $signature = 'migrate:keycloak {updated_at?}';//updated_at format =
+    protected $description = 'Copy all auth0 clients to keycloak clients - does NOT remove the auth0 clients.';
+
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    public function handle(): int
+    {
+        $updatedAt = $this->argument('updated_at');
+        $auth0Clients = $this->findAuth0Clients($updatedAt);
+
+        if ($updatedAt !== null && !preg_match('/^\d{4}-\d{2}-\d{2}$/', $updatedAt)) {
+            $this->warn('Invalid format used for updated at, use the format YYYY-MM-DD');
+            return self::FAILURE;
+        }
+
+        $total = count($auth0Clients);
+        if ($total <= 0) {
+            $this->warn($updatedAt ? 'No clients found to migrate starting from ' . $updatedAt : 'No clients found to migrate ');
+            return self::FAILURE;
+        }
+
+        if (!$this->confirm(
+            sprintf(
+                'Are you sure you want to copy %s auth0 clients to Keycloak %s?',
+                $total,
+                $updatedAt ? 'starting from ' . $updatedAt : ''
+            )
+        )) {
+            return self::FAILURE;
+        }
+
+        $bar = $this->output->createProgressBar($total);
+        $bar->start();
+
+        foreach ($auth0Clients as $client) {
+            DB::table('keycloak_clients')->insert([
+                'id' => $client->id,
+                'integration_id' => $client->integration_id,
+                'client_id' => $client->auth0_client_id,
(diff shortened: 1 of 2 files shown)


LucWollants (changes requested):
As discussed through Slack it is better to reuse the existing repositories. There is not that much added values in saving the historical time data of created keys. If needed this can also be extracted from the activity log table.

grubolsch:
Change using raw query to repository, like discussed with @LucWollants

tests/Console/Commands/Migrations/MigrateAuth0ToKeycloakClientsTest.php
```diff
@@ -0,0 +1,84 @@
...
+            'client_id' => 'auth0_client1',
+            'client_secret' => 'secret1',
+            'realm' => 'acc',
+        ]);
+    }
+
+    private function getPendingCommand(string $command, array $params = []): PendingCommand
+    {
+        $command = $this->artisan($command, $params);
+        $this->assertInstanceOf(PendingCommand::class, $command);
```
LucWollants, line 81: What is the added value of asserting this type?

tests/Console/Commands/Migrations/MigrateAuth0ToKeycloakClientsTest.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->getPendingCommand('migrate:keycloak', ['updated_at' => 'invalid-date'])
+            ->expectsOutput('Invalid format used for updated at, use the format YYYY-MM-DD')
+            ->assertExitCode(Command::FAILURE);
+    }
+
+    public function test_no_clients_to_migrate(): void
+    {
+        $this->getPendingCommand('migrate:keycloak')
+            ->expectsOutput('No clients found to migrate ')
+            ->assertExitCode(Command::FAILURE);
```
LucWollants, line 31: I think it is debatable if this is an error.

app/Console/Commands/Migrations/MigrateAuth0ToKeycloakClients.php
```diff
@@ -0,0 +1,90 @@
...
+use App\Keycloak\Client;
+use App\Keycloak\Repositories\KeycloakClientRepository;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use Ramsey\Uuid\Uuid;
+
+final class MigrateAuth0ToKeycloakClients extends Command
+{
+    protected $signature = 'migrate:keycloak {updated_at?}';//updated_at format =
```
LucWollants, line 17: Unwanted comment?

tests/Console/Commands/Migrations/MigrateAuth0ToKeycloakClientsTest.php
simon-debruijn: I would also expect this noop to result in success code

tests/Console/Commands/Migrations/MigrateAuth0ToKeycloakClientsTest.php
```diff
@@ -0,0 +1,84 @@
...
+            'auth0_client_secret' => 'secret1',
+            'auth0_tenant' => 'tenant1',
+            'created_at' => now(),
+            'updated_at' => now(),
+        ]);
+
+        // This will be updated at the start of the next century
+        $this->getPendingCommand('migrate:keycloak', ['updated_at' => '2100-01-01'])
+            ->expectsOutput('No clients found to migrate starting from 2100-01-01')
+            ->assertExitCode(Command::FAILURE);
```
simon-debruijn, line 49: Same comment as other command

simon-debruijn (approved):

tests/Console/Commands/Migrations/MigrateAuth0ToKeycloakClientsTest.php
grubolsch: artisan() can also return an int.
Phpstan gives the error

`  75     Method Tests\Console\Commands\Migrations\MigrateAuth0ToKeycloakClientsTest::getPendingCommand() should return Illuminate\Testing\PendingCommand but returns Illuminate\Testing\PendingCommand|int.  
`

I am open to alternatives. :-)

grubolsch:
Processed feedback, added extra test to test the more complex limitations of the auth0_clients query.

LucWollants (approved):

PR #1211, opened 2024-06-18
PPF-523 Deleting integration should block keycloak clients

grubolsch:
### Added
- Deleting integration now blocks keycloak clients

### Changed
BlockClients no longer fetches the integration but just uses the integration id.
- It was not needed anyway, nothing except the id is used
- It broke the delete integration flow because at that point the repository no longer finds the resource because it is deleted!

---
Ticket: https://jira.uitdatabank.be/browse/PPF-523

### DIFF
--- a/app/Keycloak/Listeners/BlockClients.php
+++ b/app/Keycloak/Listeners/BlockClients.php
@@ -6,7 +6,6 @@ namespace App\Keycloak\Listeners;
 
 use App\Domain\Integrations\Events\IntegrationBlocked;
 use App\Domain\Integrations\Events\IntegrationDeleted;
-use App\Domain\Integrations\Repositories\IntegrationRepository;
 use App\Keycloak\Client\ApiClient;
 use App\Keycloak\Exception\KeyCloakApiFailed;
 use App\Keycloak\Repositories\KeycloakClientRepository;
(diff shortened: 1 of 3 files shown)


app/Keycloak/Listeners/BlockClients.php
```diff
@@ -19,24 +19,22 @@ final class BlockClients implements ShouldQueue
...
         private readonly KeycloakClientRepository $keycloakClientRepository,
         private readonly ApiClient $client,
         private readonly LoggerInterface $logger
     ) {
     }
 
-    public function handle(IntegrationBlocked $integrationBlocked): void
+    public function handle(IntegrationBlocked|IntegrationDeleted $integrationBlocked): void
     {
-        $integration = $this->integrationRepository->getById($integrationBlocked->id);
```
LucWollants, line 31: FYI: it's fine like it is now, but there is a `getByIdWithTrashed` to get deleted integration.

app/Keycloak/Listeners/BlockClients.php
```diff
@@ -19,24 +19,22 @@ final class BlockClients implements ShouldQueue
...
     public function __construct(
-        private readonly IntegrationRepository $integrationRepository,
         private readonly KeycloakClientRepository $keycloakClientRepository,
         private readonly ApiClient $client,
         private readonly LoggerInterface $logger
     ) {
     }
 
-    public function handle(IntegrationBlocked $integrationBlocked): void
+    public function handle(IntegrationBlocked|IntegrationDeleted $integrationBlocked): void
```
LucWollants, line 28: Maybe use a more generic name for the parameter?

LucWollants (approved):

PR #1213, opened 2024-06-19
PPF-525 Changed payload create client Keycloak

grubolsch:
We need to sent a slight different payload to Keycloak to create a client, based on feedback from Erwin and Corneel.

### Added
- Add https://docs.publiq.be/ / https://publiq.stoplight.io/ to allow web origins (CORS)
- Add hidden attribute "origin" to "publiq-platform" -> this will be used later to clear test clients (discussed with Erwin)
- Set  'use.refresh.tokens' => true (discussed with Erwin)
- Added extra test to see that the IntegrationToKeycloakClientConverter & IntegrationUrlConverter play well together

### Fixed
- BUGFIX Switch Login and Callback URL, you can only have 1 login url but multiple callback urls

---
Ticket: https://jira.uitdatabank.be/browse/PPF-525

### DIFF
--- a/app/Keycloak/Converters/IntegrationUrlConverter.php
+++ b/app/Keycloak/Converters/IntegrationUrlConverter.php
@@ -4,62 +4,56 @@ declare(strict_types=1);
 
 namespace App\Keycloak\Converters;
 
+use App\Domain\Integrations\Environment;
 use App\Domain\Integrations\Integration;
 use App\Domain\Integrations\IntegrationPartnerStatus;
 use App\Domain\Integrations\IntegrationUrl;
 use App\Domain\Integrations\IntegrationUrlType;
-use App\Keycloak\Client;
 
 /* Converts integration urls in the correct Keycloak API format */
+
 final readonly class IntegrationUrlConverter
 {
-    public static function convert(Integration $integration, Client $client): array
+    public static function buildCallbackUrls(Integration $integration, Environment $environment): array
     {
-        // Empty start construct to make all urls empty at the beginning of each update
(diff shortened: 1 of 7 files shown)


app/Keycloak/Converters/IntegrationUrlConverter.php
```diff
@@ -4,62 +4,56 @@
...
+use App\Domain\Integrations\Environment;
 use App\Domain\Integrations\Integration;
 use App\Domain\Integrations\IntegrationPartnerStatus;
 use App\Domain\Integrations\IntegrationUrl;
 use App\Domain\Integrations\IntegrationUrlType;
-use App\Keycloak\Client;
 
 /* Converts integration urls in the correct Keycloak API format */
+
 final readonly class IntegrationUrlConverter
```
LucWollants, line 15: To avoid future use it is maybe best to move these methods as private methods on `IntegrationToKeycloakClientConverter`. This will prevent developers from using `IntegrationUrlConverter` directly.

grubolsch: I would prefer not to do this.
1) The argument that a developer could use the class directly in error can be made for any splitting up of code in different classes. That is the downside of the Single Responsibility Principle, and the downside of a language where you always export everything and cannot hide or seal some classes in a namespace.
2) When I started doing this I quickly realised that this would make testing several cases in IntegrationUrlConverterTest a lot more difficult. Because you can no longer target the url creation directly the short, clear defined unit tests would become really big.

LucWollants (approved):

PR #1214, opened 2024-06-19
PPF-460-UiTPAS activation

vhande:
### Added
- Organizers to the UiTPAS activation dialog

---

Ticket: https://jira.uitdatabank.be/browse/PPF-460


app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -311,6 +329,28 @@ public function storeKeyVisibilityUpgrade(string $id, KeyVisibilityUpgradeReques
         );
     }
 
+    public function getOrganizers(string $id, GetOrganizersRequest $request): RedirectResponse
```
LucWollants, line 332: I think it's better to move this to a dedicated class `OrganizerController` and change the endpoint to `GET /organizers?name={name}`

app/Domain/Integrations/FormRequests/GetOrganizersRequest.php
```diff
@@ -0,0 +1,20 @@
...
+
+final class GetOrganizersRequest extends FormRequest
+{
+    /**
+     * @return array<string, mixed>
+     */
+    public function rules(): array
+    {
+        return [
+            'organizer' => ['required', 'string'],
```
LucWollants, line 17: I would change the key to `name`

app/Domain/Integrations/Controllers/OrganizerController.php
```diff
@@ -0,0 +1,43 @@
...
+use CultuurNet\SearchV3\ValueObjects\Organizer;
+use Illuminate\Http\JsonResponse;
+
+final class OrganizerController extends Controller
+{
+    public function __construct(private readonly SearchService $searchService)
+    {
+    }
+
+    public function getOrganizers(GetOrganizersRequest $request): JsonResponse
```
Anahkiasen, line 19: Would perhaps call this class and method `OrganizersController::index` to be closer to the CRUD conventions we used for the other controllers. Or `search` perhaps? Not sure

app/Domain/Integrations/FormRequests/RequestActivationRequest.php
```diff
@@ -45,4 +52,12 @@ private function isAccountingInfoRequired(): bool
 
         return $integration->type !== IntegrationType::EntryApi || $subscription->price > 0.0;
     }
+
+    private function isUITPAS(): bool
+    {
+        /** @var IntegrationRepository $integrationRepository */
+        $integrationRepository = App::get(IntegrationRepository::class);
+        $integration = $integrationRepository->getById(Uuid::fromString($this->id));
```
Anahkiasen, line 60: Would maybe extract fetching the Integration from the two boolean check methods and pass it as argument so that it isn't done twice.

app/Domain/Integrations/Controllers/OrganizerController.php
```diff
@@ -0,0 +1,43 @@
...
+                }, $data)
+            );
+
+        } catch (\Exception $e) {
+            return new JsonResponse(
+                ['exception' => $e->getMessage()],
+                500
+            );
+            ;
+        }
```
Anahkiasen, line 41: Not sure this needs to be done manually since we have a handler for 500 errors already, what kind of exception is thrown at the moment? Or at least what happens if it's left uncatched manually at the moment?

resources/ts/Components/ActivationDialog.tsx
```diff
@@ -81,6 +84,31 @@ type Props = {
...
+type InitialValues = {
+  organization: Organization;
+  organizers: Organizer[];
+  coupon: string;
+};
+
+type Organizer = {
+  name: string;
+  id: string;
+};
```
Anahkiasen, line 110: We have a few of these in `resources/ts/types/Organization.ts` I'd perhaps move the missing ones there too and reuse them

resources/ts/Components/ActivationDialog.tsx
```diff
@@ -235,7 +299,82 @@ export const ActivationDialog = ({
...
+                        setIsSearchListVisible(false);
+                      }
+                    }}
+                    onBlur={(e) => {
+                      e.target.value = "";
+                      const timeoutId = setTimeout(() => {
+                        setIsSearchListVisible(false);
+                        clearTimeout(timeoutId);
+                      }, 200);
+                    }}
```
simon-debruijn, line 355: Can be moved into a separate function

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -290,20 +290,24 @@ public function requestActivation(string $id, RequestActivationRequest $request)
...
-        foreach ($organizers as $organizer) {
-            $organizer = new Organizer(
-                Uuid::uuid4(),
-                Uuid::fromString($id),
-                Uuid::fromString($organizer['id'])
-            );
 
-            $this->organizerRepository->create($organizer);
-        }
+        DB::transaction(function () use ($organizers, $id, $organization, $request): void {
```
LucWollants, line 295: This is not the place to put DB logic. If we want to do this in a transaction then we need to implement an extra method on the `OrganizerRepository` that saves a array of organizers.

simon-debruijn (approved):

LucWollants (approved):

PR #1215, opened 2024-06-21
PPF-450 login with keycloak

grubolsch:
### Added
- You can now login with Keycloak on both frontend and backend!

If you want to test this make sure you have the config and the certificate as described in https://github.com/cultuurnet/appconfig/pull/684

### Feedback requested
- I split up Realm to 2 classes RealmWithScopeConfig & Realm.
I not really happy how this ended up, problem is that the Client to do machine-to-machine is a different one than the login, requiring extra config. The Realm related to login has no need for Scope Config, but maybe I should have added it anyway to avoid this akward design. Very happy for ideas.

- I had lots of issues mocking "SessionManager" because the methods on it are magic methods (put, get), in the end I made a class that extends SessionManager with actual working put/get logic. I am open for suggestions.

### Related
https://github.com/cultuurnet/appconfig/pull/684

---


grubolsch:
Test fail because I did not commit the .pem file; which I don't want to do.
I have a weekend to think how to fix this...

LucWollants:
> Test fail because I did not commit the .pem file; which I don't want to do. I have a weekend to think how to fix this...

appconfig, as all other projects do?

LucWollants:
@grubolsch for `udb3-search-service` it is solved by adding some custom pem files, see https://github.com/cultuurnet/udb3-search-service/tree/main/tests/Http/Authentication/samples

PR #1223, opened 2024-06-24
PPF-450 Fix logout for keycloak

grubolsch:
### Added
- To make logout work in Keycloak we needed a different URL structure

---

### DIFF
--- a/app/Domain/Auth/Controllers/LogoutController.php
+++ b/app/Domain/Auth/Controllers/LogoutController.php
@@ -22,6 +22,11 @@ final class LogoutController
             Auth::guard(config('nova.guard'))->logout();
         }
 
+        if (env('AUTHENTICATION_MODE') === 'keycloak') {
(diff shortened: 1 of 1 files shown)


brampauwelyn (approved):

app/Domain/Auth/Controllers/LogoutController.php
```diff
@@ -22,6 +22,16 @@ private function getLogoutLink(): string
             Auth::guard(config('nova.guard'))->logout();
         }
 
+        if (env('AUTHENTICATION_MODE') === 'keycloak') {
```
LucWollants, line 25: You should use the `config` method instead of the `env` method.

app/Domain/Auth/Controllers/LogoutController.php
```diff
@@ -22,6 +22,16 @@ private function getLogoutLink(): string
...
+            return sprintf(
+                'https://%s/realms/%s/protocol/openid-connect/logout?client_id=%s&post_logout_redirect_uri=%s',
+                env('AUTH0_LOGIN_MANAGEMENT_DOMAIN'),
```
LucWollants, line 28: Mixing `AUTH0_` with `KEYCLOAK_` in a part of the code base dedicated to Keycloak is very confusing. It is also not the best approach to streamline the cleanup once Auth0 can be removed.

PR #1224, opened 2024-06-25
PPF-450 Add different config for auth0 vs keycloak

grubolsch:
### Changed
Use different env values to configure auth-flow depending if `IS_KEYCLOAK_ENABLED` is true.

### Related
https://github.com/cultuurnet/appconfig/pull/693

---

### DIFF
--- a/.env.ci
+++ b/.env.ci
@@ -74,6 +74,8 @@ AUTH0_LOGIN_CLIENT_ID=***
 AUTH0_LOGIN_CLIENT_SECRET=***
 AUTH0_LOGIN_REDIRECT_URI=http://localhost/auth/callback
 
+KEYCLOAK_LOGIN_PARAMETERS="locale=nl&referrer=publiq-platform&prompt=login&skip_verify_legacy=true&product_display_name=publiq platform"
+
 AUTH0_CLIENT_CREATION_ENABLED=true
 
 # MUST always be set to DEV tenant config except for the .env for the production app!
--- a/.env.ci
+++ b/.env.ci
@@ -1,6 +1,6 @@
 APP_NAME=publiq-platform
 APP_ENV=local
-APP_KEY=
+APP_KEY=base64:5f6ivEq1QCl8ylgkFEzMpU6npqDi5tGYRuG1LYzilb4=
 APP_DEBUG=true
 APP_URL=http://localhost
 APP_SERVICE=laravel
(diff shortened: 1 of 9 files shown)


brampauwelyn (approved):

config/auth0.php
```diff
@@ -6,30 +6,33 @@
...
     // Determines what configuration options will be required.
     'strategy' => env('AUTH0_LOGIN_STRATEGY', \Auth0\SDK\Configuration\SdkConfiguration::STRATEGY_REGULAR),
 
     // Auth0 domain for your tenant, found in your Auth0 Application settings.
-    'domain' => env('AUTH0_LOGIN_DOMAIN'),
+    'domain' => $isKeycloakEnabled ? env('KEYCLOAK_LOGIN_DOMAIN') : env('AUTH0_LOGIN_DOMAIN'),
 
     // Auth0 Management API domain for your tenant, found in your Auth0 Application settings.
-    'managementDomain' => env('AUTH0_LOGIN_MANAGEMENT_DOMAIN'),
+    'managementDomain' => $isKeycloakEnabled ? env('KEYCLOAK_LOGIN_MANAGEMENT_DOMAIN') : env('AUTH0_LOGIN_MANAGEMENT_DOMAIN'),
```
LucWollants, line 23: If Keycloak is enabled will this not break the creation of keys on Auth0? Or is this management part only used to get users?

config/auth0.php
```diff
@@ -6,30 +6,33 @@
  * Please review available configuration options here:
  * https://github.com/auth0/auth0-PHP#configuration-options.
  */
+$isKeycloakEnabled = env('IS_KEYCLOAK_ENABLED', false);
+
 return [
```
LucWollants, line 11: Maybe it is more readable to start with the if statement and "early return" an array if Keycloak login is enabled. When Keycloak is not enabled the existing Auth0 config is returned. This will also make the cleanup later more straight forward.

LucWollants: I see some tenants need to be configured always. So you could also create the login array first and merge it with the tenants part.

grubolsch: I changed the approach, i now split up the variables and change everything in the service container layer.
I also did test and all auth0 client creation still works.

app/Providers/AuthServiceProvider.php
```diff
@@ -63,14 +64,28 @@ public function boot(): void
...
         $this->app->when(LoginController::class)
             ->needs('$loginParams')
-            ->give($auth0LoginParameters);
+            ->give($this->getLoginParameters());
+    }
+
+    private function getLoginParameters(): array
+    {
+        $auth0LoginParameters = [];
+        if (config(KeycloakConfig::KEYCLOAK_LOGIN_ENABLED)) {
```
LucWollants, line 84: Maybe use an early return statement?

.env.ci
```diff
@@ -1,6 +1,6 @@
 APP_NAME=publiq-platform
 APP_ENV=local
-APP_KEY=
+APP_KEY=base64:5f6ivEq1QCl8ylgkFEzMpU6npqDi5tGYRuG1LYzilb4=
```
LucWollants, line 3: This is not needed because the CI pipeline contains this https://github.com/cultuurnet/publiq-platform/blob/main/.github/workflows/ci.yml#L69 for every step

.env.ci
```diff
@@ -74,6 +74,17 @@ AUTH0_LOGIN_CLIENT_ID=***
 AUTH0_LOGIN_CLIENT_SECRET=***
 AUTH0_LOGIN_REDIRECT_URI=http://localhost/auth/callback
 
+KEYCLOAK_CREATION_ENABLED=true
+KEYCLOAK_LOGIN_ENABLED=false
+
+KEYCLOAK_LOGIN_DOMAIN=account-keycloak-acc.uitid.be
+KEYCLOAK_LOGIN_MANAGEMENT_DOMAIN=account-keycloak-acc.uitid.be
+KEYCLOAK_LOGIN_CLIENT_ID=
+KEYCLOAK_LOGIN_CLIENT_SECRET
```
LucWollants, line 83: This is missing `=` to be consistent with the rest

PR #1226, opened 2024-06-27
Add Organizers tab to integrations

Anahkiasen:
### Added
- Add Organizers tab to integrations

---
Ticket: https://jira.publiq.be/browse/PPF-462


app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -167,11 +171,24 @@ public function show(string $id): Response
...
+        $uitpasOrganizers = $this->searchClient->findUiTPASOrganizers(...$organizerIds)->getMember()?->getItems();
+        $organizers = collect($uitpasOrganizers)->map(function (SapiOrganizer $organizer) {
+            $id = explode('/', $organizer->getId() ?? '');
+
+            return [
+                'id' => $id[count($id) - 1],
+                'name' => $organizer->getName()?->getValues() ?? $id,
+                'status' => $organizer->getWorkflowStatus() === 'ACTIVE' ? 'Live' : 'Test',
+            ];
+        });
```
Anahkiasen, line 184: @LucWollants I don't like that last part but not sure how to reconcile with the existing Organizer value object, I didn't want to make a new one for this one usecase and modifying the existing means it's hard to mark fields as required/optional for both use cases, what do you think?

LucWollants: If I understand correctly, this is to figure out if an UDB3 organizer is a test organizer or a production organizer?

There is only one fixed TEST organizer that will be used by all integrations, this organizer is "maintained" by the UiTPAS admins. All the other integrations are part of the PROD environment. Now that I think about it, there is no need to link this TEST organizer, so there is no need to store the link with the TEST organizer in the database.

Given this extra information, the best approach is maybe to just hardcode the UUID of the TEST organizer?

Anahkiasen: I was talking about the entire thing but yes the status thing is also very specific to this part of the UI. But I meant more it felt weird to build up these objects here and return plain arrays when we have an Organizer VO, I guess my question is should I create another VO specifically for what is returned from the UDB API since ours is purely about the pivot table information?

grubolsch: Here are my thoughts:

If the PHP were handling more extensive processing, I would recommend utilizing a new Value Object (VO) to enhance type safety.
However, since the data is immediately passed to the frontend where type information is inherently lost, I see no issue with passing an array in this scenario. It effectively serves the purpose of aggregating data for the view.

grubolsch (approved):

PR #1246, opened 2024-07-04
HOTFIX: Check if both auth0 & keycloak are enabled.

JonasVHG:
### Changed

- `Auth0ServiceProvider`: Check if either `Keycloak` or `auth0` is enabled to create Clients.

### Fixed

- Environments who are still on `auth0` can again create Clients.

---
Ticket: https://jira.publiq.be/browse/PPF-547

### DIFF
--- a/app/Auth0/Auth0ServiceProvider.php
+++ b/app/Auth0/Auth0ServiceProvider.php
@@ -93,7 +93,7 @@ final class Auth0ServiceProvider extends ServiceProvider
             );
         });
 
-        if (config(KeycloakConfig::KEYCLOAK_CREATION_ENABLED)) {
+        if (config(KeycloakConfig::KEYCLOAK_CREATION_ENABLED) || config('auth0.enabled')) {
             // By default, the Auth0 integration is enabled. For testing purposes this can be disabled inside the .env file.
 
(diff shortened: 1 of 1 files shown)


paulherbosch (approved):

app/Auth0/Auth0ServiceProvider.php
```diff
@@ -93,7 +93,7 @@ public function register(): void
             );
         });
 
-        if (config(KeycloakConfig::KEYCLOAK_CREATION_ENABLED)) {
+        if (config(KeycloakConfig::KEYCLOAK_CREATION_ENABLED) || config('auth0.enabled')) {
```
simon-debruijn, line 96: Does this if make sense? If it's keycloak or auth0, isn't it always true?
or is `KEYCLOAK_CREATION_ENABLED` only for a specific part e.g. management client

simon-debruijn (approved):

app/Auth0/Auth0ServiceProvider.php
JonasVHG: Yes, but on `test` & `prod` Keycloak is still `false`. So there no clients were created.
This is a temporary situation as long as test & prod are still on `auth0`.
Once every environment is on `keycloak` this can be reverted.

PR #1262, opened 2024-07-08
Added ability to tie organizer to integration

Anahkiasen:
### Added
- Added ability to tie organizer to integration

---
Ticket: https://jira.publiq.be/browse/PPF-463


app/Domain/Integrations/Organizer.php
```diff
@@ -11,7 +11,7 @@
     public function __construct(
         public UuidInterface $id,
         public UuidInterface $integrationId,
-        public UuidInterface $organizerId
+        public string $organizerId
```
JonasVHG, line 14: I wonder if the migration should also be updated for clarity? I think it still works because all the faulty UUId's have 35 instead of 36 characters? It's an open question, don't know for sure myself.

Anahkiasen: We can switch it to a standard text column but I feel I want to keep the hope we will be able to revert to actual UUIDs down the line. For now yes the `uuid` Blueprint method effectively just  creates a `char(36)` which is fine for our use case so not sure either 😬

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -292,12 +293,26 @@ public function updateOrganization(string $id, UpdateOrganizationRequest $reques
...
     }
 
+    public function updateOrganizers(string $integrationId, UpdateIntegrationOrganizersRequest $request): RedirectResponse
+    {
+        $integration = $this->integrationRepository->getById(Uuid::fromString($integrationId));
+
+        $organizerIds = collect($integration->organizers())->map(fn (Organizer $organizer) => $organizer->organizerId);
+        $newOrganizers = array_filter(OrganizerMapper::map($request, $integrationId), static function (Organizer $organizer) use ($organizerIds) {
+            return !in_array($organizer->organizerId, $organizerIds->toArray(), true);
+        });
```
simon-debruijn, line 303:
```suggestion
        $newOrganizers = array_filter(
          OrganizerMapper::map($request, $integrationId),
          fn (Organizer $organizer) => !in_array($organizer->organizerId, $organizerIds->toArray(), true)
        );
```

This removes a bit of syntax overhead

app/Domain/Integrations/Mappers/OrganizerMapper.php
```diff
@@ -13,7 +14,7 @@ final class OrganizerMapper
     /**
      * @return Organizer[]
      */
-    public static function map(RequestActivationRequest $request, string $id): array
+    public static function map(RequestActivationRequest|UpdateIntegrationOrganizersRequest $request, string $id): array
```
simon-debruijn, line 17: On other Mappers we added a mapUpdate for example and it passes it to map. Then you can only pass 1 type.
Either approach is fine to me

app/Domain/Integrations/Mappers/OrganizerMapper.php
```diff
@@ -24,7 +25,7 @@ public static function map(RequestActivationRequest $request, string $id): array
             $organizers[] = new Organizer(
                 Uuid::uuid4(),
                 Uuid::fromString($id),
-                Uuid::fromString($organizer['id'])
+                $organizer['id']
```
simon-debruijn, line 28: This is a string right?

simon-debruijn: Never mind, I see why

resources/ts/Components/Integrations/Detail/OrganizersDatalist.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const [organizerList, setOrganizerList] = useState<UiTPASOrganizer[]>([]);
+  const [organizerError, setOrganizerError] = useState(false);
+  const organizersInputRef = useRef<HTMLInputElement>(null);
+
+  const handleGetOrganizers = debounce(
+    async (e: React.ChangeEvent<HTMLInputElement>) => {
+      const response = await fetch(`/organizers?name=${e.target.value}`);
+      try {
+        const data = await response.json();
+        if ("exception" in data) {
```
simon-debruijn, line 29:
```suggestion
        if (!!data && typeof data === "object" && "exception" in data) {
```

Will throw an error on null otherwise

You should probably handle `if !data` seperatly

resources/ts/Components/Integrations/Detail/OrganizersDatalist.tsx
```diff
@@ -0,0 +1,152 @@
...
+        );
+        setOrganizerList(organizers);
+      } catch (error) {
+        setOrganizerError(true);
+        return;
+      }
+
+      if (organizerError) {
+        setOrganizerError(false);
+      }
```
simon-debruijn, line 49: if error set error to false?

resources/ts/Components/Integrations/Detail/OrganizersDatalist.tsx
```diff
@@ -0,0 +1,152 @@
...
+        {...props}
+        component={
+          <>
+            <Input
+              type="text"
+              name="organizers"
+              ref={organizersInputRef}
+              onChange={async (e) => {
+                await handleInputOnChange(e);
+              }}
```
simon-debruijn, line 128:
```suggestion
              onChange={(e) => handleInputOnChange(e)}
```
If eslint allows it

simon-debruijn (changes requested):

resources/ts/Components/Integrations/Detail/OrganizersDatalist.tsx
Anahkiasen: It clears out any error from the past attempt if the logic managed to reach this point, since actual errors will exit out of the logic before it

simon-debruijn (approved):

PR #1263, opened 2024-07-09
PPF-553 BUGFIX Organizers gives 409 in admin panel

grubolsch:
### Fixed
The UiTdatabank organizers linked to an UiTPAS integration do not load in the admin panel. (409 error).
The problem is that on this API call the $resource is not loaded in the IntegrationModel (aka the fields).
This means that the method isUitpas() always returns false, because the $type is not set.

As a (quick) fix I set the default value of type to Uitpas.
On the UI overview pages the Integration is loaded in, so this overwrites the default value.

Know problem: When creating an integration in the Admin UI the Type is as a default Uitpas

Known Limitation: If we ever have 2 different values in type that toggle conditions, we have a problem.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-553


JonasVHG (approved):
This is oke a hotfix, but we should really look into a more robust solution.

PR #1265, opened 2024-07-09
PPF-555 For uitpas integration always use key visibility v2

grubolsch:
### Changed
For UiTPAS API integration, the key visiblity must always be v2
As discussed with Corneel, this is only changed for Integrations created in the frontend.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-555

### DIFF
--- a/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
+++ b/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
@@ -561,4 +561,28 @@ final class EloquentIntegrationRepositoryTest extends TestCase
             'partner_status' => IntegrationPartnerStatus::FIRST_PARTY,
         ]);
     }
+
+    // https://jira.publiq.be/browse/PPF-555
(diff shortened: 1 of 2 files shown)


tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
```diff
@@ -561,4 +561,28 @@ public function test_it_can_save_partner_status(): void
...
+        $searchIntegration = new Integration(
+            $integrationId,
+            IntegrationType::UiTPAS,
+            'First party integration',
+            'First party integration',
+            Uuid::uuid4(),
+            IntegrationStatus::Draft,
+            IntegrationPartnerStatus::FIRST_PARTY,
+        );
+        $searchIntegration = $searchIntegration->withKeyVisibility(KeyVisibility::v1);
```
simon-debruijn, line 579: I'm wondering if this should maybe throw an exception instead of magically setting the visibility differently than expected

simon-debruijn (approved):

PR #1267, opened 2024-07-10
PPF-459 uitpas integration activation

JonasVHG:
### Added

- `ActivateUitpasIntegration`: Action to activate an UiTPAS-integration with 1 or more Organizers.
- `Resources/Integration`: Injected `ActivateUitpasIntegration` into Integration

### Fixed

- An admin can now activate an Integration with 1 or More ID's of an `Organizer`

---
Ticket: https://jira.publiq.be/browse/PPF-459

### DIFF
--- a/app/Nova/Actions/ActivateUitpasIntegration.php
+++ b/app/Nova/Actions/ActivateUitpasIntegration.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Nova\Actions;
+
+use App\Domain\Integrations\Models\IntegrationModel;
+use App\Domain\Integrations\Organizer;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Domain\Integrations\Repositories\OrganizerRepository;
+use App\Domain\Organizations\Models\OrganizationModel;
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Support\Collection;
+use Laravel\Nova\Actions\Action;
+use Laravel\Nova\Actions\ActionResponse;
+use Laravel\Nova\Fields\ActionFields;
+use Laravel\Nova\Fields\Select;
+use Laravel\Nova\Fields\Text;
+use Laravel\Nova\Http\Requests\NovaRequest;
+use Ramsey\Uuid\Uuid;
+
+final class ActivateUitpasIntegration extends Action
+{
+    use InteractsWithQueue;
+    use Queueable;
+
+    public function __construct(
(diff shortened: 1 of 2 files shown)


app/Nova/Actions/ActivateUitpasIntegration.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public function handle(ActionFields $fields, Collection $integrations): ActionResponse
+    {
+        /** @var IntegrationModel $integration */
+        $integration = $integrations->first();
+
+        /** @var string $organizationIdAsString */
+        $organizationIdAsString = $fields->get('organization');
+        $organizationId = Uuid::fromString($organizationIdAsString);
```
grubolsch, line 41: make   $organizationIdAsString inline, and you can drop the docblock

app/Nova/Actions/ActivateUitpasIntegration.php
```diff
@@ -0,0 +1,85 @@
...
+
+        /** @var string $organizers */
+        $organizers = $fields->get('organizers');
+        $organizerArray = array_map('trim', explode(',', $organizers));
```
grubolsch, line 45: You could make organizers inline, and then do the following:

```
        $organizers = array_map('trim', explode(',', $fields->get('organizers')));

        foreach ($organizers as $organizer) {
```

app/Nova/Actions/ActivateUitpasIntegration.php
```diff
@@ -0,0 +1,85 @@
...
+    public function __construct(
+        private readonly IntegrationRepository $integrationRepository,
+        private readonly OrganizerRepository $organizerRepository
+    ) {
+    }
+
+    public function handle(ActionFields $fields, Collection $integrations): ActionResponse
+    {
+        /** @var IntegrationModel $integration */
+        $integration = $integrations->first();
```
grubolsch, line 37: Are we sure we will always have an Integration?
Otherwise I would run a check with an instanceof.
That is also a way to avoid the docblock.

grubolsch (approved):
2 minor remarks

app/Nova/Actions/ActivateUitpasIntegration.php
JonasVHG: Since it's an action on an Integration, there should always be at least 1 Integration.

app/Nova/Actions/ActivateUitpasIntegration.php
JonasVHG: Maybe best not to do this is PHPStorm complains about it, it's the same in other Action(s). So maybe just leave it as it is?

PR #1310, opened 2024-08-06
PPF-562 Create helper method to activate uitpas integration with organizers

grubolsch:
### Added
- Added new unit tests for UdbOrganizerRepository, class was untested.
- Added new check in EloquentIntegrationRepositoryTest to make sure Organizers are created in db.
- Added new Collection for UdbOrganizer
- 
### Changed
- Added 2 versions of UdbOrganizerRepository::create(), the new one is called "createInBulk", the later one wraps all inserts in a transaction.
- requestActivation and activateWithOrganization now wrap the creation of the Organizer.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-562

### DIFF
--- a/app/Domain/Integrations/Controllers/IntegrationController.php
+++ b/app/Domain/Integrations/Controllers/IntegrationController.php
@@ -334,7 +334,7 @@ final class IntegrationController extends Controller
             Uuid::fromString($id),
             $organization->id,
             $request->input('coupon'),
-            new UdbOrganizers(...UdbOrganizerMapper::mapActivationRequest($request, $id))
+            UdbOrganizerMapper::mapActivationRequest($request, $id)
         );
 
         return Redirect::back();
--- a/app/Domain/Integrations/Controllers/IntegrationController.php
+++ b/app/Domain/Integrations/Controllers/IntegrationController.php
@@ -37,7 +37,7 @@ use App\Domain\Integrations\UdbOrganizer;
 use App\Domain\Integrations\Repositories\IntegrationRepository;
 use App\Domain\Integrations\Repositories\IntegrationUrlRepository;
 use App\Domain\Integrations\Repositories\UdbOrganizerRepository;
-use App\Domain\Integrations\UdbOrganizerCollection;
+use App\Domain\Integrations\UdbOrganizers;
 use App\Domain\KeyVisibilityUpgrades\KeyVisibilityUpgrade;
 use App\Domain\KeyVisibilityUpgrades\Repositories\KeyVisibilityUpgradeRepository;
 use App\Domain\Organizations\Repositories\OrganizationRepository;
@@ -304,7 +304,7 @@ final class IntegrationController extends Controller
             fn (UdbOrganizer $organizer) => !in_array($organizer->organizerId, $organizerIds->toArray(), true)
         );
 
-        $this->organizerRepository->createInBulk(new UdbOrganizerCollection($newOrganizers));
+        $this->organizerRepository->createInBulk(new UdbOrganizers($newOrganizers));
 
         return Redirect::back();
     }
@@ -334,7 +334,7 @@ final class IntegrationController extends Controller
             Uuid::fromString($id),
             $organization->id,
             $request->input('coupon'),
-            new UdbOrganizerCollection(...UdbOrganizerMapper::mapActivationRequest($request, $id))
+            new UdbOrganizers(...UdbOrganizerMapper::mapActivationRequest($request, $id))
         );
 
         return Redirect::back();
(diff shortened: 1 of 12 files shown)


app/Domain/Integrations/UdbOrganizerCollection.php
```diff
@@ -0,0 +1,14 @@
...
+declare(strict_types=1);
+
+namespace App\Domain\Integrations;
+
+use Illuminate\Support\Collection;
+
+/**
+ * @extends Collection<int, UdbOrganizer>
+ */
+final class UdbOrganizerCollection extends Collection
```
LucWollants, line 12: For other collections we used a naming convention to just use the plural form. For example `Realm` and `Realms`, so for this case it should be `UdbOrganizers`

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
```diff
@@ -115,15 +129,25 @@ public function activate(UuidInterface $id): void
         $integrationModel->activate();
     }
 
-    public function activateWithOrganization(UuidInterface $id, UuidInterface $organizationId, ?string $couponCode): void
+    public function activateWithOrganization(UuidInterface $integrationId, UuidInterface $organizationId, ?string $couponCode, array $organizers=[]): void
```
LucWollants, line 132: Should this also be typed on the new collection?

app/Domain/Integrations/Repositories/EloquentUdbOrganizerRepository.php
```diff
@@ -6,11 +6,21 @@
...
+    public function create(UdbOrganizer $organizer): void
+    {
+        UdbOrganizerModel::query()->create([
+            'id' => $organizer->id->toString(),
+            'integration_id' => $organizer->integrationId->toString(),
+            'organizer_id' => $organizer->organizerId,
+        ]);
+    }
+
+    public function createInBulk(UdbOrganizerCollection $organizers): void
```
LucWollants, line 23: You can reuse the `create` method inside the `createInBulk` method

app/Domain/Integrations/Repositories/IntegrationRepository.php
```diff
@@ -17,8 +18,8 @@ public function getById(UuidInterface $id): Integration;
     public function getByIdWithTrashed(UuidInterface $id): Integration;
     public function deleteById(UuidInterface $id): ?bool;
     public function getByContactEmail(string $email, ?string $searchQuery): PaginatedCollection;
-    public function requestActivation(UuidInterface $id, UuidInterface $organizationId, ?string $couponCode): void;
+    public function requestActivation(UuidInterface $id, UuidInterface $organizationId, ?string $couponCode, UdbOrganizerCollection $organizers=null): void;
     public function activate(UuidInterface $id): void;
-    public function activateWithOrganization(UuidInterface $id, UuidInterface $organizationId, ?string $couponCode): void;
+    public function activateWithOrganization(UuidInterface $integrationId, UuidInterface $organizationId, ?string $couponCode, array $organizers=[]): void;
```
LucWollants, line 23: Why change a single parameter? Is it not better to keep it consistent with the rest of the methods?

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
grubolsch: I thought so to at first but if you look at ActivateIntegration you can see that actually it is an array of strings (ids).
I will add a docblock to make this more clear.

grubolsch:
Feedback has been processed

app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -329,10 +330,12 @@ public function requestActivation(string $id, RequestActivationRequest $request)
...
 
-        $organizers = UdbOrganizerMapper::mapActivationRequest($request, $id);
-        $this->organizerRepository->create(...$organizers);
-
-        $this->integrationRepository->requestActivation(Uuid::fromString($id), $organization->id, $request->input('coupon'));
+        $this->integrationRepository->requestActivation(
+            Uuid::fromString($id),
+            $organization->id,
+            $request->input('coupon'),
+            new UdbOrganizers(...UdbOrganizerMapper::mapActivationRequest($request, $id))
```
LucWollants, line 337: Is it possible to make `mapActivationRequest` return `UdbOrganizers`? (I forgot to press "add comment" for this one)

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
LucWollants: But should it be an array of strings? The `UdbOrganizers` could be created where this gets called. So in the `handle` method of `ActivateUitpasIntegration` the construction of `UdbOrganizer` could be done.

LucWollants (approved):

app/Domain/Integrations/Controllers/IntegrationController.php
LucWollants: I think it returns a single organization, so there is no need for a collection here.

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
grubolsch: Moved it to the ActivateUitpasIntegration class

PR #1329, opened 2024-08-20
PPF-53 Add basis for sending transactional emails

grubolsch:
### Added
- Added the basis for sending transactional e-mails with Mailjet.

---


app/Domain/Mail/MailNotSend.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Domain\Mail;
+
+final class MailNotSend extends \DomainException
```
LucWollants, line 7: Maybe add the ID of the integration for which it failed?

LucWollants (approved):

app/Domain/Mail/MailNotSend.php
grubolsch: While adding the id is difficult because i don't have an integration in that context (nor do I want to add it because not every mail is necessarly related to an integration), I now add the variables in the exception.
This should provide extra logging.

PR #1331, opened 2024-08-20
PPF-55 Send mail when integration is blocked

grubolsch:
### Added
Sent mail when integration is blocked

---
Ticket: https://jira.uitdatabank.be/browse/PPF-55

### DIFF
--- a/app/Domain/Mail/MailManager.php
+++ b/app/Domain/Mail/MailManager.php
@@ -2,13 +2,12 @@
 
 declare(strict_types=1);
 
-namespace App\Mails;
+namespace App\Domain\Mail;
 
 use App\Domain\Contacts\Contact;
 use App\Domain\Integrations\Events\IntegrationActivated;
 use App\Domain\Integrations\Events\IntegrationBlocked;
 use App\Domain\Integrations\Repositories\IntegrationRepository;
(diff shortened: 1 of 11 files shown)


app/Domain/Mail/MailManager.php
```diff
@@ -37,14 +39,35 @@ public function sendIntegrationActivatedMail(IntegrationActivated $integrationAc
...
                     'integrationName' => $integration->name,
-                    //@todo Should we add a language preference to contacts?
                     'url' => $this->baseUrl . '/nl/integraties/' . $integration->id,
                     'type' => $integration->type->value,
                 ]
             );
         }
     }
 
+    public function sendIntegrationBlockedMail(IntegrationBlocked $integrationBlocked): void
```
LucWollants, line 49: Is the direct URL of the integration inside `publiq-platform` useful for every type of mail?

LucWollants (approved):

app/Domain/Mail/MailManager.php
grubolsch: Difficult to say because we don't have a final copy of these emails.
Currently I am just using it for the activation email, but we will have to see how it ends up.
Would you prefer to always send all variables?

PR #1333, opened 2024-08-20
PPF-570 - Make e2e tests less flaky

simon-debruijn:
### Changed
- Decrease workers due too high load
- Augment certain timeouts
- Never return page from factory function
- Move assertion url into factory functions (so we are sure this is always done)

### Fixed
- Make url assertions more correct

---
Ticket: https://jira.uitdatabank.be/browse/PPF-570

### DIFF
--- a/e2e/setup/auth.setup.ts
+++ b/e2e/setup/auth.setup.ts
@@ -31,7 +31,6 @@ setup("authenticate as admin", async ({ page }) => {
 
   await page.getByRole("button", { name: "Meld je aan", exact: true }).click();
 
-  await page.waitForLoadState("networkidle");
   await page.waitForURL("/nl/integraties");
 
   await page.context().storageState({ path: "playwright/.auth/admin.json" });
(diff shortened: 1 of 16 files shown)


vhande (approved):

LucWollants:
I did several runs locally and I can confirm that it now works. Also the commands from the `Makefile` are working.

As discussed through Slack one improvement could be to create contacts on the detail page of an integration instead of the contacts resources page.

simon-debruijn:
> I did several runs locally and I can confirm that it now works. Also the commands from the `Makefile` are working.
> 
> As discussed through Slack one improvement could be to create contacts on the detail page of an integration instead of the contacts resources page.

Done https://github.com/cultuurnet/publiq-platform/pull/1333/commits/a0f8785deda42a310e4891c72ab4ab2cc4145dec

LucWollants (approved):
Great work on making the E2E tests reliable!

PR #1357, opened 2024-09-02
PPF-587 Import UiTPAS integrations

LucWollants:
### Added
- Added script to migrate UiTPAS integrations from CSV file. This will import the basic integration details, contacts and keys

---
Ticket: https://jira.uitdatabank.be/browse/PPF-587

### DIFF
--- a/app/Console/Commands/UiTPAS/MigrateUiTPAS.php
+++ b/app/Console/Commands/UiTPAS/MigrateUiTPAS.php
@@ -4,6 +4,10 @@ declare(strict_types=1);
 
 namespace App\Console\Commands\UiTPAS;
 
+use App\Auth0\Auth0Client;
+use App\Auth0\Auth0ClusterSDK;
+use App\Auth0\Auth0Tenant;
+use App\Auth0\Repositories\Auth0ClientRepository;
 use App\Console\Commands\ReadCsvFile;
 use App\Domain\Contacts\Repositories\ContactRepository;
 use App\Domain\Integrations\Integration;
@@ -32,6 +36,8 @@ final class MigrateUiTPAS extends Command
         private readonly IntegrationRepository $integrationRepository,
         private readonly SubscriptionRepository $subscriptionRepository,
         private readonly ContactRepository $contactRepository,
+        private readonly Auth0ClusterSDK $auth0ClusterSDK,
+        private readonly Auth0ClientRepository $auth0ClientRepository
     ) {
         parent::__construct();
     }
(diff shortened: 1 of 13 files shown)


app/Console/Commands/UiTPAS/UiTPASIntegration.php
```diff
@@ -0,0 +1,110 @@
...
+use App\Auth0\Auth0Tenant;
+use App\Domain\Contacts\Contact;
+use App\Domain\Contacts\ContactType;
+use App\Domain\Integrations\IntegrationStatus;
+use App\Domain\Integrations\Website;
+use InvalidArgumentException;
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+
+final class UiTPASIntegration
```
grubolsch, line 16: I would change this DTO object in the following ways:
- In the constructor expect all the specifiek fields (id, email, lastname) as separate arguments, and give them their own properties.
- Have a method "fromArray" to accepts the current array and splits up all the fields.

This would have several benefits:
- Would make this class more reusable for other situations / cleaner to mock
- Gives you type hinting for all the separate fields
- Makes the code more readable because you only have the indexes on one place and don't have to remember "index 2 is the email", etc...

app/Console/Commands/UiTPAS/MigrateUiTPAS.php
```diff
@@ -0,0 +1,144 @@
...
+        $rows = $this->readCsvFile('database/uitpas/integrations.csv');
+        $uitpasIntegrations = array_map(
+            fn (array $row) => new UiTPASIntegration($row),
+            $rows
+        );
+
+        $integrationsCount = count($uitpasIntegrations);
+        if ($integrationsCount <= 0) {
+            $this->warn('No UiTPAS integrations to import');
+            return 0;
```
grubolsch, line 58: Return self::SUCCESS instead

app/Console/Commands/UiTPAS/MigrateUiTPAS.php
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        if (!$this->confirm('Are you sure you want to import ' . $integrationsCount . ' UiTPAS integrations?')) {
+            return 0;
```
grubolsch, line 62: Return self::SUCCESS instead

app/Console/Commands/UiTPAS/MigrateUiTPAS.php
```diff
@@ -0,0 +1,144 @@
...
+
+            $this->migrateContacts($integrationId, $uitpasIntegration);
+
+            $this->migrateAuth0Clients($integrationId, $uitpasIntegration);
+
+            $this->info($integrationId . ' - Ended importing project ' . $uitpasIntegration->name());
+            $this->info('---');
+        }
+
+        return 0;
```
grubolsch, line 80: Return self::SUCCESS instead

grubolsch (changes requested):

app/Console/Commands/UiTPAS/UiTPASIntegration.php
LucWollants: The goal of this class is to wrap the row of the CSV file, if I would change it to accept typed parameters there is no longer a need for this class. I could rename it to `UiTPASIntegrationFromCSV` or something like that.

PR #1365, opened 2024-09-06
PPF-54 Send welcome mail

grubolsch:
### Added
Send a welcome e-mail when a user creates an integration.

### Fixed
Getting the base url for the email variable was broken.

### FAQ
- Why did I not use the existing event IntegrationCreated?
When this event is thrown the contacts are not attached to it yet, meaning I don't know to who to sent the e-mail.

- The email is not sent when creating an integration from the admin.
I know, same problem as before, because the flow is different in the admin you can create an integration without any contact. I asked Corneel if we should provide a workaround for this. (See ticket)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-54

### DIFF
--- a/app/Domain/Integrations/Controllers/IntegrationController.php
+++ b/app/Domain/Integrations/Controllers/IntegrationController.php
@@ -10,7 +10,7 @@ use App\Domain\Contacts\ContactType;
 use App\Domain\Contacts\Repositories\ContactKeyVisibilityRepository;
 use App\Domain\Contacts\Repositories\ContactRepository;
 use App\Domain\Coupons\Repositories\CouponRepository;
-use App\Domain\Integrations\Events\IntegrationCreatedPost;
+use App\Domain\Integrations\Events\IntegrationCreatedAfter;
 use App\Domain\Integrations\FormRequests\KeyVisibilityUpgradeRequest;
 use App\Domain\Integrations\FormRequests\RequestActivationRequest;
 use App\Domain\Integrations\FormRequests\StoreContactRequest;
--- a/app/Domain/Integrations/Controllers/IntegrationController.php
+++ b/app/Domain/Integrations/Controllers/IntegrationController.php
@@ -10,6 +10,7 @@ use App\Domain\Contacts\ContactType;
 use App\Domain\Contacts\Repositories\ContactKeyVisibilityRepository;
 use App\Domain\Contacts\Repositories\ContactRepository;
 use App\Domain\Coupons\Repositories\CouponRepository;
+use App\Domain\Integrations\Events\IntegrationCreatedPost;
 use App\Domain\Integrations\FormRequests\KeyVisibilityUpgradeRequest;
 use App\Domain\Integrations\FormRequests\RequestActivationRequest;
 use App\Domain\Integrations\FormRequests\StoreContactRequest;
@@ -131,6 +132,8 @@ final class IntegrationController extends Controller
         } else {
             $this->integrationRepository->save($integration);
         }
+
+        IntegrationCreatedPost::dispatch($integration->id);
 
         return Redirect::route(
             TranslatedRoute::getTranslatedRouteName($request, 'integrations.show'),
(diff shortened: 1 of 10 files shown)


app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -132,6 +133,8 @@ public function store(StoreIntegrationRequest $request): RedirectResponse
             $this->integrationRepository->save($integration);
         }
 
+        IntegrationCreatedPost::dispatch($integration->id);
```
LucWollants, line 136: This could also be dispatched inside the repository to have it in a more common/general place?

LucWollants: Maybe you could dispatch it inside `EloquentIntegrationRepository::saveTransaction`

app/Domain/Integrations/Events/IntegrationCreatedPost.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Domain\Integrations\Events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use Ramsey\Uuid\UuidInterface;
+
+final class IntegrationCreatedPost
```
LucWollants, line 10: What does `Post` mean in this context?

grubolsch: "After". I will take suggestions on this name. :-)

grubolsch:
> > When this event is thrown the contacts are not attached to it yet, meaning I don't know to who to sent the e-mail.
> 
> There is a solution for this, which is also implemented inside `SyncWidget`
> 
> This listener has more events it handles, it listens to the `IntegrationCreated` event and the `ContactCreated` event. Only when all integration data and all contacts are present send the welcome mail.
> 
> Sending a welcome mail has a special requirement, the mail should only be sent once. Not for every `ContactCreated` event even after the first mail was send.

Do I understand correctly that you would listen to `ContactCreated` instead of my new event?
I thought of that but thought it would be weird if you add a contact for an integration that already exists for a long time and you get a welcome e-mail.

app/Domain/Integrations/Controllers/IntegrationController.php
grubolsch: done

app/Domain/Integrations/Events/IntegrationCreatedPost.php
grubolsch: Renamed it to `IntegrationCreatedAfter`

LucWollants: That is still a name that doesn't make much sense. You want to make clear that the contacts are also created, so something like `IntegrationCreatedWithContacts` can be fired from the repository.

LucWollants:
> Do I understand correctly that you would listen to `ContactCreated` instead of my new event?

No, both `IntegrationCreated` and `ContactCreated`

>I thought of that but thought it would be weird if you add a contact for an integration that already exists for a long time and you get a welcome e-mail.

Sending a welcome mail has a special requirement, the mail should only be sent once. Not for every ContactCreated event even after the first mail was sent.

grubolsch:
I don't think I get it.
There would be no point to listen to `IntegrationCreated`, there will never be a situation where in that scenario the contacts are already available.
If I listen to `ContactCreated` I can sent the e-mail, one at a time, but there is no concept of "only sent it once", because each Contact would only trigger ContactCreated once. I can not for example put a flag in Integration because then only 1 contact will recieve an email.

app/Domain/Integrations/Events/IntegrationCreatedPost.php
grubolsch: Oké, I renamed it

LucWollants:
> There would be no point to listen to `IntegrationCreated`, there will never be a situation where in that scenario the contacts are already available.

These events are placed on queues so in theory the `ContactCreated` could be handled before `IntegrationCreated`. But the straightforward approach with `IntegrationCreatedWithContacts` is fine for this use case.

I mainly wanted to point out that there are situations where the listener has to make the final check if an event needs to be handled or not. In this case of sending the mail, a table that keeps track of integration which already got a welcome mail could solve this.

LucWollants (approved):

PR #1377, opened 2024-09-11
PPF-57 Sent reminder email

grubolsch:
### Added
- Sent a mail to integrators when after a year the integration is still not activated.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-57


grubolsch:
Feedback has been processed

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
```diff
@@ -0,0 +1,62 @@
...
+namespace App\Console\Commands;
+
+use App\Domain\Integrations\IntegrationType;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Mails\MailJet\MailjetConfig;
+use App\Mails\SendIntegrationActivationReminderEmail;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+final class SendIntegrationActivationReminderEmailCommand extends Command
```
LucWollants, line 14: I think the suffix `Command` is not really needed.

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    protected $signature = 'cronjob:send-activation-reminder-email {--force : Skip confirmation prompt}';
```
LucWollants, line 16: Is it better to put this in the `integration:` category instead of `cronjob:`?

LucWollants:
I wonder if the requirements change after this is implemented: https://jira.publiq.be/browse/PPF-564
Maybe we need to avoid sending emails to integrations that were put on hold by an admin.

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function handle(): int
+    {
+        if (!config(MailjetConfig::TRANSACTIONAL_EMAILS_ENABLED)) {
+            $this->output->writeln('Email feature flag is disabled - mails not sent');
+            return self::FAILURE;
+        }
+
+        $integrationTypeConfig = [
+            IntegrationType::SearchApi->value => 6,
```
LucWollants, line 35: What are the 6 and 3? If those are the IDs of the Mailjet templates, then it may be worth putting them in an enum.

app/Domain/Integrations/Integration.php
```diff
@@ -53,7 +54,8 @@ public function __construct(
         public readonly string $description,
         public readonly UuidInterface $subscriptionId,
         public readonly IntegrationStatus $status,
-        public readonly IntegrationPartnerStatus $partnerStatus
+        public readonly IntegrationPartnerStatus $partnerStatus,
+        public ?DateTimeInterface $reminderEmailSent = null,
```
LucWollants, line 58: This should not be part of the default constructor because it has a `with` method.

app/Domain/Integrations/Models/IntegrationModel.php
```diff
@@ -340,6 +343,7 @@ public function toDomain(): Integration
             Uuid::fromString($this->subscription_id),
             $this->status,
             $this->partner_status,
+            $this->reminder_email_sent ? Carbon::parse($this->reminder_email_sent) : null
```
LucWollants, line 346: Use the `with` method only when it's not null

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
```diff
@@ -50,9 +52,24 @@ public function update(Integration $integration): void
             'status' => $integration->status,
             'partner_status' => $integration->partnerStatus,
             'key_visibility' => $integration->getKeyVisibility(),
+            'reminder_email_sent' => $integration->reminderEmailSent,
         ]);
     }
 
+    /** @return Collection<Integration> */
+    public function getIntegrationsThatHaveNotBeenActivatedYetByType(IntegrationType $type, int $months): Collection
```
LucWollants, line 60: I think something like `getDraftsByTypeAndMonths` covers the implementation better.

app/Mails/SendIntegrationActivationReminderEmail.php
```diff
@@ -0,0 +1,46 @@
...
+use App\Domain\Contacts\Contact;
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Domain\Mail\MailManager;
+use Carbon\Carbon;
+use Illuminate\Console\OutputStyle;
+use Illuminate\Support\Collection;
+use Psr\Log\LoggerInterface;
+
+final readonly class SendIntegrationActivationReminderEmail
```
LucWollants, line 16: Shouldn't this be a task to execute on the queue? Otherwise, the command to send all the mails will block per mail

app/Mails/SendIntegrationActivationReminderEmail.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    public function __construct(
+        private MailManager $mailManager,
+        private IntegrationRepository $integrationRepository,
+        private LoggerInterface $logger
+    ) {
+    }
+
+    public function send(Collection $integrations, OutputStyle $output): void
```
LucWollants, line 25: Feels a bit weird to inject something console-specific into the domain layer.

app/Mails/SendIntegrationActivationReminderEmail.php
LucWollants: Now that I have checked the existing implementation, I don't think the `MailManager` gets queued. It does get triggered by an event but it is missing the specific traits for queueing like `use Queueable;`

database/migrations/2024_09_11_095233_add_sent_reminder_mail_flag_to_integration.php
LucWollants, line 1: Match the name of the column also in the file name?

tests/Console/Commands/SendIntegrationActivationReminderEmailCommandTest.php
```diff
@@ -0,0 +1,46 @@
...
+use App\Mails\MailJet\MailjetConfig;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Testing\PendingCommand;
+use Tests\TestCase;
+
+final class SendIntegrationActivationReminderEmailCommandTest extends TestCase
+{
+    use RefreshDatabase;
+
+    public function test_feature_flag_is_off(): void
```
LucWollants, line 17: This method name should be more explicit. Something like `test_it_warns_about_disabled_transactional_mails`

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
```diff
@@ -698,6 +702,89 @@ public function test_it_can_not_save_integration_with_subscription_with_differen
         );
     }
 
+    /** @dataProvider dataProviderGetIntegrationsThatHaveNotBeenActivatedYet */
+    public function test_get_integrations_that_have_not_been_activated_yet(IntegrationType $integrationType, IntegrationStatus $status, Carbon $date, ?Carbon $reminderEmailSent, bool $hasContact, int $expectedCount): void
```
LucWollants, line 706: Can this be put on multiple lines?

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
```diff
@@ -698,6 +702,89 @@ public function test_it_can_not_save_integration_with_subscription_with_differen
...
+        $this->assertCount($expectedCount, $this->integrationRepository->getIntegrationsThatHaveNotBeenActivatedYetByType(IntegrationType::SearchApi, 12));
+    }
+
+    public static function dataProviderGetIntegrationsThatHaveNotBeenActivatedYet(): array
+    {
+        return [
+            'Should not be selected: wrong type' => [
+                IntegrationType::EntryApi,
+                IntegrationStatus::Active,
+                Carbon::now()->subYears(2),
```
LucWollants, line 740: This will fail after a certain time, you can use `Carbon::setTestNow` or even more advanced testing features explained here https://laravel.com/docs/11.x/mocking#interacting-with-time

app/Domain/Mail/MailManager.php
```diff
@@ -18,13 +18,15 @@ final class MailManager
...
     private const SUBJECT_INTEGRATION_CREATED = 'Welcome to Publiq platform - Let\'s get you started!';
+    private const SUBJECT_INTEGRATION_ACTIVATION_REMINDER_EMAIL = 'Publiq platform - Can we help you to activate your integration?';
 
     public function __construct(
         private readonly Mailer $mailer,
         private readonly IntegrationRepository $integrationRepository,
         private readonly int $templateIntegrationCreated,
         private readonly int $templateIntegrationActivated,
         private readonly int $templateIntegrationBlocked,
+        private readonly int $templateIntegrationActivationReminderEmail,
```
LucWollants, line 29: I think this can move to a class that holds the template IDs, otherwise with every new template you need to change the signature of this method.

LucWollants (changes requested):
Mostly naming remarks, but also an important one about the queueing behaviour.

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
grubolsch: Those are after how many months we sent an email. 
I made constants out of them.

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
grubolsch: Renamed it to "getDraftsByTypeAndOlderThenMonthsAgo" to communicate that it looks at integrations X months ago.

app/Domain/Mail/MailManager.php
grubolsch: I got exactly that idea, but I wanted to do it in a seperate pr if that is oké for you?

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
grubolsch: I am familiar with the theory behind time in unit tests and Carbon, I use it in SendIntegrationActivationReminderEmailTest, but I do not think it is a problem here.

I create a fresh integration each time with the time of "now", and then I create a date X months/years ago, and then I directly check it with a query.
I do not need to directly assert this date, and the test does not work with a fixed date.

Am I overlooking something?

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
grubolsch: It needs to have this because the service it uses it called `SendIntegrationActivationReminderEmail`.
I open to suggestions for another name.

grubolsch:
> I wonder if the requirements change after this is implemented: https://jira.publiq.be/browse/PPF-564 Maybe we need to avoid sending emails to integrations that were put on hold by an admin.

I am going to ask this to Corneel.

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
```diff
@@ -32,8 +35,8 @@ public function handle(): int
         }
 
         $integrationTypeConfig = [
-            IntegrationType::SearchApi->value => 6,
-            IntegrationType::Widgets->value => 3,
+            IntegrationType::SearchApi->value => self::SEARCH_API_SEND_MAIL_AFTER_X_MONTHS,
```
LucWollants, line 38: Is that not something typically for an enum?

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
LucWollants: Maybe call the service `SendActivationReminder`. This also makes me wonder of there is not too much redundancy in the naming, it is always for `Integration` and it is always an `Email`

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
LucWollants: You are correct, it will work in the future because a new time is generated for every provided test.

app/Domain/Mail/MailManager.php
LucWollants: Fine for me to do this in a separate pull request.

app/Mails/SendIntegrationActivationReminderEmail.php
grubolsch: Made them Queueable, also made a ticket for infra to configure the cronjobs once my code is delivered. https://jira.publiq.be/browse/PPF-601

app/Mails/SendIntegrationActivationReminderEmail.php
```diff
@@ -0,0 +1,44 @@
...
+        private MailManager $mailManager,
+        private IntegrationRepository $integrationRepository,
+        private LoggerInterface $logger
+    ) {
+    }
+
+    public function send(Collection $integrations): void
+    {
+        foreach ($integrations as $integration) {
+            $this->mailManager->sendActivationReminderEmail(new ActivationReminderEmailSend($integration->id));
```
LucWollants, line 28: This is not how `Events` are supposed to be handled. They should be dispatched and the listener will be called by the framework, not with direct method calls.

app/Domain/Integrations/Events/ActivationReminderEmailSend.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Domain\Integrations\Events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use Ramsey\Uuid\UuidInterface;
+
+final class ActivationReminderEmailSend
```
LucWollants, line 10: This event indicates that the Reminder mail was sent, but it is the trigger to send them, so maybe something like `ActivationExpired` matches it better

app/Console/Commands/SendIntegrationActivationReminderEmailCommand.php
grubolsch: Renamed class to SearchExpiredIntegrations, and removed the old Service Dependency (like discussed on slack) because it didn't do anything anymore.

The handle() now throws an Event called ActivationExpired and all references to sending an email are gone.

Also add a new test for the new way of working.

grubolsch:
> I wonder if the requirements change after this is implemented: https://jira.publiq.be/browse/PPF-564 Maybe we need to avoid sending emails to integrations that were put on hold by an admin.

Made a comment in the other ticket and discussed this with Corneel. Good remark!

grubolsch:
All feedback has been processed

app/Domain/Integrations/Integration.php
```diff
@@ -203,13 +205,25 @@ public function keycloakClients(): array
...
+    }
+
     public function withUrls(IntegrationUrl ...$urls): self
     {
         $clone = clone $this;
         $clone->urls = $urls;
         return $clone;
     }
 
+    public function withreminderEmailSent(DateTimeInterface $reminderEmailSent): self
```
LucWollants, line 220:
```suggestion
    public function withReminderEmailSent(DateTimeInterface $reminderEmailSent): self
```

app/Domain/Integrations/Integration.php
```diff
@@ -53,7 +55,7 @@ public function __construct(
         public readonly string $description,
         public readonly UuidInterface $subscriptionId,
         public readonly IntegrationStatus $status,
-        public readonly IntegrationPartnerStatus $partnerStatus
+        public readonly IntegrationPartnerStatus $partnerStatus,
```
LucWollants, line 58:
```suggestion
        public readonly IntegrationPartnerStatus $partnerStatus
```

app/Domain/Integrations/IntegrationType.php
```diff
@@ -10,4 +10,15 @@ enum IntegrationType: string
     case SearchApi = 'search-api';
     case Widgets = 'widgets';
     case UiTPAS = 'uitpas';
+
+    public static function fromName(string $name): self
```
LucWollants, line 14: The new `enum` class has a built-in method to do this, see https://www.php.net/manual/en/class.backedenum.php

app/Domain/Integrations/Integration.php
grubolsch: I actually choose on purpose to leave these trailing comma's in.
The idea is that if we ever add a new option you would have not to change anything on the previous line, meaning that in a PR that line would not show up making the change more clear.
Happy to change it if it is not in line with the current coding standards, but it does have a purpose.

app/Domain/Integrations/IntegrationType.php
grubolsch: Are you talking about "from()"? That matches on value, not on name like I have.
And the reason I have to use this is because the value comes from another Enum, IntegrationExpirationTime, which has the name not the value as case name.
And the reason I do not use the value there is because these contain dashes (eg. entry-api) making them invalid case names.

app/Domain/Integrations/Integration.php
LucWollants: We have enforced this rule with trailing commas on arrays but not on methods. If we want to enforce it, we could add a rule for this.

app/Domain/Integrations/IntegrationType.php
LucWollants: Now that I read the code again, this is maybe just the conversion from `IntegrationExpirationTime` to `IntegrationType`. That also made me wonder, as discussed through Slack, if the `IntegrationExpirationTime` doesn't miss entries for `UiTPAS` and `EntryAPI`

app/Domain/Integrations/Integration.php
grubolsch: Happy to create a rule for it later. Is it oké if I let it be for now?

app/Domain/Integrations/IntegrationType.php
grubolsch: Removed `IntegrationExpirationTime` and moved it into config.
Also makes it possible to drop the akward conversion of `IntegrationExpirationTime` to `IntegrationType`

app/Domain/Integrations/Integration.php
LucWollants: Sure

config/mailjet.php
```diff
@@ -12,6 +13,12 @@
         'key' => env('MAILJET_API_KEY'),
         'secret' => env('MAILJET_API_SECRET'),
     ],
+    'expiration_timers' => [ // Timer is always in months
```
LucWollants, line 16: Nice approach!

LucWollants (approved):
Thanks for all the refactors. We have a very readable and consistent implementation now for mails, which follows the same flow as for other external tools like Insightly and Keycloak.

PR #1390, opened 2024-09-17
PPF-53 Move template config to seperate class

grubolsch:
### Changed
- Moved all config for the different mail templates to a separate class (Templates), so we don't get an ever growing list of properties in the constructor of MailManager.
- Added the ability to disable specific mails in the mailjet config.
- Currently the blocked email is disabled because after talking to Corneel about this, it depends on another feature.

---


app/Domain/Mail/MailManager.php
```diff
@@ -147,4 +93,26 @@ private function getUniqueContactsWithPreferredContactType(Integration $integrat
 
         return $uniqueContacts;
     }
+
+    public function sendMail(MailEnabled $event, ?Template $template): ?Integration
```
LucWollants, line 97: Inject the `$id` here so there is no need for the interface `MailEnabled`.

There is also no reason for `$template` to be nullable. That indicates a misconfiguration and should throw.

app/Domain/Mail/MailManager.php
```diff
@@ -147,4 +93,26 @@ private function getUniqueContactsWithPreferredContactType(Integration $integrat
...
+            $this->mailer->send(
+                $this->getFrom(),
+                $this->getAddresses($contact),
+                $template->id,
+                $template->subject,
+                $this->getIntegrationVariables($contact, $integration)
+            );
+        }
+
+        return $integration;
```
LucWollants, line 116: Is this return required?

grubolsch: Yes, because for `sendActivationReminderEmail()` I need it to update the ReminderEmailSent.
I wanted to avoid fetching and building the `Integration` twice.

grubolsch: Nevermind, going to change it

app/Mails/Template/Templates.php
```diff
@@ -0,0 +1,29 @@
...
+namespace App\Mails\Template;
+
+use Illuminate\Support\Collection;
+
+/**
+ * @extends Collection<string, Template>
+ */
+final class Templates extends Collection
+{
+    public const INTEGRATION_CREATED = 'integration_created';
```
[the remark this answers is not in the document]
grubolsch: done, it is called "templateName"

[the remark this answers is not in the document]
LucWollants: Can these constants be removed now?

config/mailjet.php
```diff
@@ -3,17 +3,35 @@
...
     'api' => [
         'key' => env('MAILJET_API_KEY'),
         'secret' => env('MAILJET_API_SECRET'),
     ],
-    'templates' => [
-        'integration_created' => env('MAILJET_TEMPLATE_INTEGRATION_CREATED'),
-        'integration_blocked' => env('MAILJET_TEMPLATE_INTEGRATION_BLOCKED'),
-        'integration_activated' => env('MAILJET_TEMPLATE_INTEGRATION_ACTIVATED'),
-        'integration_activation_reminder' => env('MAILJET_TEMPLATE_INTEGRATION_ACTIVATION_REMINDER'),
+    'mails' => [
```
LucWollants, line 15: Is this not config for `templates` instead of `mails`?

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
```diff
@@ -60,7 +60,9 @@ public function update(Integration $integration): void
     public function getDraftsByTypeAndOlderThenMonthsAgo(IntegrationType $type, int $months): Collection
     {
         return IntegrationModel::query()
+            ->distinct()
             ->where('status', 'draft')
+            ->where('type', $type->value)
```
LucWollants, line 65: Weird that this was not covered with a unit test, should the unit test be updated and include different integration types? I think the test "Should not be selected: wrong type" is still wrong because it is an active integration

grubolsch: Yes, the status in the test was the problem. It has been updated.

app/Mails/Template/Templates.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Illuminate\Support\Collection;
+use InvalidArgumentException;
+
+/**
+ * @extends Collection<string, Template>
+ */
+final class Templates extends Collection
+{
+    public static function build(array $mails): self
```
LucWollants, line 15: Nit: maybe use `$templates` as the name here? I see a mix of 3 names now, the class is called `Templates`, the input parameter is called `$mails` and a single item is called `$config`

LucWollants (approved):
Thanks for the extra refactors.

PR #1392, opened 2024-09-18
PPF-53 Added whitelist for sandbox mode

grubolsch:
### Added
Expanded the sandbox mode logic.
You now have an extra .env variable with a whitelist for email domains or full email addresses. These emails will always be sent for real, even if sandbox mode is enabled.

---


config/mailjet.php
```diff
@@ -2,12 +2,12 @@
 
 declare(strict_types=1);
 
-
 use App\Mails\Template\TemplateName;
 
 return [
     'enabled' => env('MAILJET_TRANSACTIONAL_EMAILS_ENABLED', false),
     'sandbox_mode' => env('MAILJET_SANDBOX_MODE', true),
+    'sandbox_allowed_domains' => env('MAILJET_SANDBOX_ALLOWED_DOMAINS', []),
```
LucWollants, line 10: This can now return two different types. When `MAILJET_SANDBOX_ALLOWED_DOMAINS` is provided it will return a comma-separated string, when it is not provided it will return an empty array. To avoid this I would put the conversion from string to array here.

tests/Mails/MailJet/SandboxHelperTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+        $this->assertSame($expected, $sandboxHelper->getSandboxMode($to));
+    }
+
+    public static function sandboxModeProvider(): array
+    {
+        $allowedDomains = ['publiq.be', 'smurfen@gmail.com'];
+
+        return [
+            'Sandbox mode is off, should return false' => [false, $allowedDomains, ['test@gmail.com'], false],
```
LucWollants, line 29: This is rather hard to read when everything is packed together on one line.

LucWollants: I think you also missed a test for when the sandbox mode is off, but the email matches the allowed domains. I know the internal implementation and I know this doesn't make a difference, but it's a valid case we need to prevent side effects from a refactor.

app/Mails/MailJet/SandboxHelper.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Mails\MailJet;
+
+use App\Domain\Mail\Addresses;
+
+final readonly class SandboxHelper
```
LucWollants, line 9: You could name this class just `SandboxMode` and call the method `forAddresses`

app/Mails/MailJet/MailjetMailer.php
```diff
@@ -17,15 +17,15 @@
...
+        private SandboxHelper $sandboxHelper,
     ) {
 
     }
 
     public function send(Address $from, Addresses $to, int $templateId, string $subject, array $variables = []): void
     {
         $body = [
-            'SandboxMode' => $this->sandboxMode,
+            'SandboxMode' => $this->sandboxHelper->getSandboxMode($to),
```
LucWollants, line 28: I think there is a potential risk here! The `$to` is a list of email addresses, when one of them is an allowed domain, all the other addresses also receive an email.

LucWollants (changes requested):
I think there is a logical error, when one of the to addresses matches the domain, all the other addresses will also get a mail.

app/Mails/MailJet/MailjetMailer.php
grubolsch: Currently we always sent a mail one at at time because we sent as variables the firstname/lastname of each contact.
So for now this is not a problem.
So this would only be a problem for future changes, I think we have 3 options here)

1) Keep it as is but put a comment on the class/function warning for this flaw.
2) Block all emails
3) More difficult - Manipulate the Addresses to change the collection to only contain valid emails. Not a fan of this because of the complexity, we don't need this currently.

What would you prefer?

LucWollants: I see a fourth option: only disable the sandbox mode when all email addresses match the allowed domains. Putting a comment will be to tricky, this could result in sending mails to real integrators.

grubolsch: Went for option 5

LucWollants (approved):
Takes for the extra refactor. It is very easy to follow the flow of events now. I guess this qualifies as less is more 😊

PR #1395, opened 2024-09-19
PPF-55 Remove the blocked email

grubolsch:
### Removed
No plans for the forseeable future to have a mail when an integration is blocked so let us remove the code.
In case of blocking an integration the servicedesk/partner cell will always contact the integrator before the event.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-55


LucWollants (changes requested):
I think the binding should also be removed https://github.com/cultuurnet/publiq-platform/blob/main/app/Mails/MailServiceProvider.php#L60

Maybe it's also good to remove it from the config if it is no longer used https://github.com/cultuurnet/appconfig/blob/main/files/platform/docker/env#L45

LucWollants (approved):

PR #1405, opened 2024-09-23
Maintenance page for Keycloak migration

grubolsch:
### Added
Added a Maintenance page for the Keycloak migration.
Screenshot 2024-09-23 at 10 51 22

### DIFF
--- a/public/503.php
+++ b/public/503.php
@@ -1,0 +1,69 @@
+<?php
+header('HTTP/1.1 503 Service Temporarily Unavailable');
+header('Status: 503 Service Temporarily Unavailable');
+?><!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>503 | Service Unavailable</title>
+
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha1/dist/css/bootstrap.min.css" rel="stylesheet">
+
+    <style>
+        body {
+            background-color: #F5F5F5;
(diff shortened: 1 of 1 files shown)


public/503.php
```diff
@@ -0,0 +1,69 @@
...
+            background-color: #009FDE;
+            color: white;
+            padding: 15px;
+            text-align: center;
+        }
+    </style>
+</head>
+<body>
+<div class="container">
+    <img src="/build/assets/Hero-DagNCfZ2.svg" alt="Maintenance Image" class="header-image">
```
LucWollants, line 56: I don't see this asset `Hero-DagNCfZ2.svg`

LucWollants:
For now, this hardcoded approach is fine, but it should move to this folder https://github.com/cultuurnet/publiq-platform/tree/main/resources/views/errors

public/503.php
grubolsch: It already exists, I think it gets deployed there with NPM.
I just copied the path from the homepage.

LucWollants: Ah it is `Hero.svg`

LucWollants (approved):
Feel free to use this for now, but afterwards refactor this to a React page like the 404 page.

grubolsch:
> For now, this hardcoded approach is fine, but it should move to this folder https://github.com/cultuurnet/publiq-platform/tree/main/resources/views/errors
I tried that, but the problem is that Inertia and the default Laravel code are not playing nice on this one.
I put files in the views/errors but that didn't do anything, in fact it crashed because it didn't have a resources/views/app.blade.php file.
When I added that, It would show that template, but not the inner templates in errors/. I could see that I tried to load the entire app.blade.php file in some JS which failed in the browser (Inertia?)

All of these things are probably solvable, but I decided it was not worth it for this short time.

PR #1407, opened 2024-09-23
BUGFIX When configured env variables arrive as string, should be int

grubolsch:
### Fixed
Related to the transactional email feature, when the months for the expiration email are configured, it throws an error because env() returns the values as string, but I expect them as integer.


JonasVHG (commented):
I think it might also be a good idea to add `MAILJET_EXPIRATION_TIMER_*` values to `.env.ci`?

config/mailjet.php
```diff
@@ -14,10 +14,10 @@
         'secret' => env('MAILJET_API_SECRET'),
     ],
     'expiration_timers' => [ // Timer is always in months
-        IntegrationType::EntryApi->value => env('MAILJET_EXPIRATION_TIMER_ENTRY_API', 6),
-        IntegrationType::SearchApi->value => env('MAILJET_EXPIRATION_TIMER_SEARCH_API', 6),
-        IntegrationType::UiTPAS->value => env('MAILJET_EXPIRATION_TIMER_UITPAS', 6),
-        IntegrationType::Widgets->value => env('MAILJET_EXPIRATION_TIMER_WIDGETS', 3),
+        IntegrationType::EntryApi->value => (int)env('MAILJET_EXPIRATION_TIMER_ENTRY_API', 6),
```
LucWollants, line 17: I'm not sure if this is the best place to fix this, Laravel has built-in support for this https://laravel.com/docs/11.x/configuration#accessing-configuration-values

It is advised to use `Config::integer`

grubolsch: :-O O, that is very good to know, I will change it!

grubolsch: Oké, now I look into it, while this is good knowledge, I am afraid I cannot use it here.
I don't directly access each variable with Config::get(), I pass the entire array.

This gives me a lot more flexibility for in the future if for example we have to not sent mails for one specific type, or quickly add a type.
I would like to keep it this way.

Alternatively I can do a cast to int in app/Console/Commands/SearchExpiredIntegrations.php if you would prefer this.

LucWollants: Indeed valid remark! It's fine to just keep the `int` conversion on the `env` function

LucWollants (approved):

JonasVHG (approved):

PR #1409, opened 2024-09-23
PPF-608 remove key secret

JonasVHG:
### Added

- `migrations/2024_09_23_103549_remove_legacy_consumer_secret_from_uitidv1_consumers.php`: Add migration to remove legacy `consumer_secret`, which is never used & should not be shown.

### Changed

- `UiTiDv1/UiTiDv1Consumer`: Remove parameter `consumerSecret` from constructor.
- `MigrateProjects` & `UiTiDv1EnvironmentSDK`:  updated constructors of `UiTiDv1Consumer`
- `EloquentUiTiDv1ConsumerRepository`: Do not save `consumer_secret`
- `UiTiDv1ConsumerModel`: Remove `consumer_secret`
- `Nova/Resources/UiTiDv1`: Remove `consumer_secret` & no longer show `consumer_key`

### Fixed

- `consumer_secret` is no longer stored, used or show in the admin panel.
- `consumer_key` is no longer shown in the admin panel, it is still necessary to save it to open an Integration in `UiTIDv1`

---
Ticket: https://jira.publiq.be/browse/PPF-608


LucWollants (changes requested):
Maybe we should just hide the columns in Nova for now. Once UiTiD v1 is no longer in use we can remove the full domain folder.

JonasVHG:
To avoid potential problems in the future, we will not remove any UiTIDv1 functionality in this PR and it will be closed.
A new PR with only cosmetic changes will be made.

PR #1414, opened 2024-09-24
Added dialog to request Slack access with custom email

Anahkiasen:
### Added
- Added dialog to request Slack access with custom email

---
Ticket: https://jira.publiq.be/browse/PPF-607


app/Http/Controllers/SupportController.php
```diff
@@ -24,11 +24,13 @@ public function index(Request $request): Response
             'slackStatus' => $slackStatus,
         ]);
     }
+
     public function sendInvitation(Request $request): RedirectResponse
     {
         $botToken = config('slack.botToken');
         $channelID = config('slack.channels.technical_support');
-        $email = Auth::user()?->email;
+        $data = $request->validate(['email' => ['required', 'email']]);
```
LucWollants, line 32: When doing the validation to check for the presence of `email` I wonder if it makes sense to have the fallback to the current user on the next line.

Anahkiasen: Do you mean merge this line and the one below?

LucWollants: The fallback to `Auth::user()?->email` will only be triggered if the `email` field is optional or not set, but in this case, you’ve marked it as `required` in the validation rules.

PR #1415, opened 2024-09-24
PPF-611 Change reminder email to only sent mail under 1 year old

grubolsch:
### Changed
In preparation for ticket https://jira.publiq.be/browse/PPF-605, we only want to sent the first reminder email for integrations younger then 1 year, to prevent sending 2 mails at the same time to an integrator.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-611

### DIFF
--- a/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
+++ b/tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
@@ -735,7 +735,10 @@ final class EloquentIntegrationRepositoryTest extends TestCase
             ]);
         }
 
-        $this->assertCount($expectedCount, $this->integrationRepository->getDraftsByTypeAndOlderThenMonthsAgo(IntegrationType::SearchApi, 12));
+        $this->assertCount(
+            $expectedCount,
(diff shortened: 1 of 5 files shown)


tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
```diff
@@ -776,15 +779,23 @@ public static function dataProviderGetIntegrationsThatHaveNotBeenActivatedYet():
...
+                Carbon::now()->subMonths(14),
                 Carbon::now(),
                 true,
                 0,
             ],
+            'Should not be selected: Too old' => [
+                IntegrationType::SearchApi,
+                IntegrationStatus::Draft,
+                Carbon::now()->subMonths(50),
+                Carbon::now(),
```
LucWollants, line 791: I think this test data is not completely correct. The mail was sent for this one, so you are not sure that the 50 months are taken into account.

LucWollants (changes requested):
I don't think the test data is correct for the new between clause.

tests/Domain/Integrations/Repositories/EloquentIntegrationRepositoryTest.php
grubolsch: Updated the test case.

LucWollants (approved):

PR #1417, opened 2024-09-24
HOTFIX - cache is too agressive and caches only the ACC master realm key

grubolsch:
### Fixed
On test / prod Keycloak is broken because the cache for the token caches based on the realm name. But both test/acc/prod have a realm called "master", so it only cached the first, which was the acc env.

This fixes this by adding the url to the cache key.

### DIFF
--- a/app/Keycloak/TokenStrategy/ClientCredentials.php
+++ b/app/Keycloak/TokenStrategy/ClientCredentials.php
@@ -28,7 +28,7 @@ final class ClientCredentials implements TokenStrategy
 
     public function fetchToken(KeycloakHttpClient $client, Realm $realm): string
     {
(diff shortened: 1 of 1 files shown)


app/Keycloak/TokenStrategy/ClientCredentials.php
```diff
@@ -28,7 +28,7 @@ public function __construct(
 
     public function fetchToken(KeycloakHttpClient $client, Realm $realm): string
     {
-        $key = $realm->internalName . $realm->clientId;
+        $key = $realm->baseUrl . $realm->internalName . $realm->clientId;
```
LucWollants, line 31: Using just the `$environment` seems more readable?

LucWollants (approved):

PR #1419, opened 2024-09-25
PPF-605 Move integration mails to separate table

grubolsch:
### Changed
Before I kept if the reminder email was sent in a column in the "integrations" table.
However in preparation for ticket PPF-605, I will need to sent and track and second e-mail.
If I follow the same way of developing I might end up with many extra columns in the "integrations" table.

In order to prevent this I create a new table called "integrations_emails", that keeps for which integration what type of email has been sent when.

This will make the future development and tracking of emails easier, needed for this ticket. 

---
Ticket: https://jira.uitdatabank.be/browse/PPF-605


app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
```diff
@@ -47,31 +48,44 @@ public function update(Integration $integration): void
...
+    public function getDraftsByTypeAndBetweenMonthsOld(IntegrationType $type, int $startMonths, int $endMonths, string $mailType): Collection
     {
         return IntegrationModel::query()
             ->distinct()
             ->where('status', 'draft')
             ->where('type', $type->value)
-            ->whereNull('reminder_email_sent')
             ->whereBetween('created_at', [Carbon::now()->subMonths($endMonths), Carbon::now()->subMonths($startMonths)])
             ->has('contacts')  // This ensures that only integrations with at least one contact are returned
+            ->whereNotExists(function (QueryBuilder $query) use ($mailType) {
```
LucWollants, line 68: I think you could add a relation to the `IntegrationModel` this could make this query easier to build and maintain. There is also a feature called `scopes` in Laravel. I think it's worth to have a look, even if not applicable to this situation.

app/Domain/Integrations/Repositories/IntegrationRepository.php
```diff
@@ -33,5 +34,6 @@ public function activateWithOrganization(UuidInterface $id, UuidInterface $organ
     public function approve(UuidInterface $id): void;
 
     /** @return Collection<Integration> */
-    public function getDraftsByTypeAndBetweenMonthsOld(IntegrationType $type, int $startMonths, int $endMonths): Collection;
+    public function getDraftsByTypeAndBetweenMonthsOld(IntegrationType $type, int $startMonths, int $endMonths, string $mailType): Collection;
```
LucWollants, line 37: Here is a new variable name `mailType` but I think that is just the template name?

database/migrations/2024_09_25_102533_create_integrations_mails_table.php
```diff
@@ -0,0 +1,33 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('integrations_mails', static function (Blueprint $table) {
+            $table->uuid('integration_id')->primary();
+            $table->string('type');
```
LucWollants, line 14: Is this the integration type or the template name?

grubolsch: Renamed it to "template_name".

app/Domain/Integrations/Repositories/EloquentIntegrationRepository.php
grubolsch: I added a relation and added scopes instead of the whereNotExists() check.

app/Domain/Integrations/Repositories/IntegrationRepository.php
grubolsch: I removed "mailType" and made it pass everywhere the enum TemplateName, I also call the variable this.

JonasVHG (approved):

PR #1420, opened 2024-09-25
PPF-605 Change cli command

grubolsch:
### Changed
Instead of 1 mail after 3/6 months, we are going to sent 2 reminder mails to integrations, the second one after 1 year.

### Related
Config changes: https://github.com/cultuurnet/appconfig/pull/828

---
Ticket: https://jira.uitdatabank.be/browse/PPF-605


app/Domain/Integrations/Events/ActivationExpired.php
```diff
@@ -11,7 +11,7 @@ final class ActivationExpired
 {
     use Dispatchable;
 
-    public function __construct(public readonly UuidInterface $id)
+    public function __construct(public readonly UuidInterface $id, public readonly string $templateName)
```
LucWollants, line 14: Maybe make the full class readonly?

Is there a reason to pass the string value instead of the enum type here?

app/Console/Commands/SearchExpiredIntegrations.php
```diff
@@ -15,50 +15,91 @@
 final class SearchExpiredIntegrations extends Command
 {
     private const ONE_YEAR = 12;
-    protected $signature = 'integration:search-expired-integrations {--force : Skip confirmation prompt}';
+    private const ONE_CENTURY = 12000; // Just a very long time to fetch all integrations older than 1 year
+
+    protected $signature = 'integration:search-expired-integrations
+                        {--force : Skip confirmation prompt}
+                        {--only-first-warning : Only sent the first warning}
```
LucWollants, line 22: Here warning is used, but that is maybe reminder?

app/Mails/MailJet/MailjetConfig.php
```diff
@@ -14,4 +14,5 @@ final class MailjetConfig
     public const SANDBOX_ALLOWED_DOMAINS = 'mailjet.sandbox_allowed_domains';
     public const MAILJET_TEMPLATES = 'mailjet.templates';
     public const MAILJET_EXPIRATION_TIMERS = 'mailjet.expiration_timers';
+    public const MAILJET_EXPIRATION_TIMERS_FINAL_WARNING = 'mailjet.expiration_timers_final_warning';
```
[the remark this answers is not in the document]
grubolsch: Made it say consistently "reminder". Sounds more friendly.

app/Domain/Integrations/Events/ActivationExpired.php
LucWollants: The difference between an enum and a string is not that big. There is also a feature in Laravel that supports this, see https://laravel.com/docs/11.x/queues#class-structure

grubolsch:
processed feedback, also added an extra test for the new final reminder email.

JonasVHG (approved):

PR #1421, opened 2024-09-26
PPF-613 Migrated keycloaks not visible

grubolsch:
### Added
Added log for full URI each time Keycloak api is called. I often missed this during debugging.

### Fixed
If a keycloak client in Platform cannot be found in Keycloak itself, instead of showing no keycloak clients at all I just give it the status "not in sync". (There might be a root issue for out of sync but that is outside scope here).

---
Ticket: https://jira.uitdatabank.be/browse/PPF-613


app/Nova/ActionGuards/Keycloak/BlockKeycloakClientGuard.php
```diff
@@ -21,6 +22,10 @@ public function canDo(object $client): bool
             return false;
         }
 
-        return ! $this->cachedKeycloakClientStatus->isClientBlocked($client);
+        try {
+            return ! $this->cachedKeycloakClientStatus->isClientBlocked($client);
+        } catch (KeyCloakApiFailed) {
+            return false;
```
LucWollants, line 28: Is it not weird to convert an exception here to a boolean? This means it wouldn't be logged in Sentry and it feels like "everything" is normal. I think it's even fine we had the errors inside the admin panel, it was an indication that something was wrong. Preferably those issues are logged with an error and will arrive in Sentry.

LucWollants:
I'm not convinced of the added value of this change. When there is an issue we should not "hide" it, we should fix it.

app/Nova/ActionGuards/Keycloak/BlockKeycloakClientGuard.php
LucWollants: I wonder if the best approach is to make the combination of integration id and environment unique in the table and still keep the uncaught exception? With that restriction, the initial migration would have failed.

grubolsch:
After a discussion Luc and I agreed on the following approach:
- UI should not fail to load
- But if Keycloak is down it should give a warning in sentry

Might need to look into the package sentry/sentry-laravel for the sentry error

config/logging.php
```diff
@@ -119,6 +119,11 @@
         'emergency' => [
             'path' => storage_path('logs/laravel.log'),
         ],
+
+        'sentry' => [
+            'driver' => 'sentry',
+            'level' => 'critical',
```
LucWollants, line 125: Is it meaningful to also log errors?

LucWollants (approved):

config/logging.php
grubolsch: We could do that - I wanted to start conservative because I didn't know what box of pandora I would unleash with error. Lots of code has not been written with the idea the error() would go into sentry, I don't know if it would be good.
My instinct says no for now, so there is a functional difference between throwing critiical and error, but if you prefer I add the error level also fine.

LucWollants: Why would that open a box of Pandora? An `error` is something that prevents the normal behaviour of the application and should be fixed. That would mean publiq-platform often fails without us knowing it 😅

grubolsch: Now I see we are talking about Sentry, not slack errors, yeah, it makes sense to log that, I will change it.

PR #1447, opened 2024-10-07
PPF-621 Change integrations mails to include all mails

grubolsch:
Currently only the mail for reminder and final reminder are inserted in the integrations_mails table.

For future logging purposes it would be useful if all transactional emails where logged in this table.

### Changed
- Add random id column to table, make this the primary key.
- Add an index on (integration_id, template_name)
- Add entry into the table integrations_mails for all transactional emails in their handlers.

### Removed
Some minor boy scouting changes
- Remove "date" from IntegrationMailRepository (this field was forgotten to be removed)
- Remove "Subject" from tests because subjects are now configured in Mailjet (was in previous pr)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-621

### DIFF
--- a/app/Domain/Integrations/IntegrationMail.php
+++ b/app/Domain/Integrations/IntegrationMail.php
@@ -10,6 +10,7 @@ use Ramsey\Uuid\UuidInterface;
 final readonly class IntegrationMail
 {
     public function __construct(
+        public UuidInterface $uuid,
         public UuidInterface $integrationId,
         public TemplateName $templateName,
     ) {
--- a/app/Domain/Integrations/IntegrationMail.php
+++ b/app/Domain/Integrations/IntegrationMail.php
@@ -7,11 +7,11 @@ namespace App\Domain\Integrations;
 use App\Mails\Template\TemplateName;
 use Ramsey\Uuid\UuidInterface;
 
-final class IntegrationMail
+final readonly class IntegrationMail
 {
     public function __construct(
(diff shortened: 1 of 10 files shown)


app/Domain/Integrations/IntegrationMail.php
```diff
@@ -7,11 +7,12 @@
 use App\Mails\Template\TemplateName;
 use Ramsey\Uuid\UuidInterface;
 
-final class IntegrationMail
+final readonly class IntegrationMail
 {
     public function __construct(
-        public readonly UuidInterface $integrationId,
-        public readonly TemplateName $templateName,
+        public UuidInterface $uuid,
```
LucWollants, line 13: I think other classes use `id` for this.

app/Mails/Template/Template.php
```diff
@@ -7,9 +7,9 @@
 final class Template
 {
     public function __construct(
-        public string $type,
+        public TemplateName $name,
         public int $id,
-        public bool $enabled,
+        public bool $enabled = true,
```
LucWollants, line 12: Was there a good reason to give this a default value?

LucWollants (approved):
I already approved, the remarks are minor.

app/Mails/Template/Template.php
LucWollants: I prefer a reason related to some added business values instead of easier tests. So if you don't mind that would to great yes.

PR #1451, opened 2024-10-08
PPF-622 Switch email validation to filter var

grubolsch:
### Changed
Email validation now used the filter_var() function (wrapped by Laravel).
Our email validation was on RFC level, but mailjet required email in a more strict format.
This means that before can could create an integration with an email like "t@t", which will pass form validation but then crash on mailjet.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-622


LucWollants:
What about the extra options here https://laravel.com/docs/11.x/validation#rule-email?

app/Mails/MailJet/MailjetMailer.php
```diff
@@ -23,6 +23,12 @@ public function __construct(
 
     public function send(Address $from, Address $to, int $templateId, array $variables = []): void
     {
+        if (!filter_var($from->getAddress(), FILTER_VALIDATE_EMAIL)) {
```
LucWollants, line 26: This is again an example of converting a valid runtime error to something that silently continues.

LucWollants (changes requested):
Nice solution with the `filter` option for validation, but I think we should not just ignore runtime errors.

app/Mails/MailJet/MailjetMailer.php
grubolsch: This issue should not occur with new integrations due to the updated email validation process. The fix is created for backwards compatibility with old data.

Currently email reminders fail when encountering an integration with invalid data, requiring manual fixes for the email addresses.

One alternative is to silence all errors from the Mailjet API, but this would prevent distinguishing between a failure caused by an invalid email and other API issues. As a result, we wouldn’t be able to confirm if the email was actually sent, only that an attempt was made to call the API.

LucWollants: Catch the errors and log them?

grubolsch: I removed it

LucWollants (approved):

PR #1465, opened 2024-10-16
Fixed blank UI for integrations with no keycloak client

Anahkiasen:
### Fixed
- Fixed blank UI for integrations with no keycloak client

---

@simon-debruijn @grubolsch Some integrations do not yet have a Keycloak client to show, even if the related env variables are enabled. As it is my understanding that this is an "unhappy path" state, and that otherwise we should always have one to show, I've tried to simplify the logic to trust the data first and foremost in the meantime, so that users aren't too confused. But since the previous condition relied on `KEYCLOAK_TEST_CLIENT_ENABLED` – which from my understanding is a temporary thing – I'm a bit unsure what the actual real condition should be if that's not enough. Especially the isLive part. So all that to say I tried to trust the data in place but perhaps it's shortsighted.

---
Ticket: https://jira.publiq.be/browse/PPF-624

### DIFF
--- a/resources/ts/Components/IntegrationClientCredential.tsx
+++ b/resources/ts/Components/IntegrationClientCredential.tsx
@@ -49,7 +49,8 @@ export const IntegrationClientCredentials = ({
   const { config } = usePageProps();
 
   const clientWithLabels =
-    config.keycloak.enabled || (!isLive && config.keycloak.testClientEnabled)
+    config.keycloak.enabled ||
+    (!isLive && config.keycloak.testClientEnabled && keycloakClient)
       ? [
           {
             label: "details.credentials.client_id",
--- a/resources/ts/Components/IntegrationClientCredential.tsx
+++ b/resources/ts/Components/IntegrationClientCredential.tsx
@@ -6,6 +6,7 @@ import { IntegrationStatus } from "../types/IntegrationStatus";
 import { Alert } from "./Alert";
 import type { Integration } from "../types/Integration";
 import type { AuthClient, KeycloakClient } from "../types/Credentials";
+import { usePageProps } from "../hooks/usePageProps";
 
 type Props = {
   client: {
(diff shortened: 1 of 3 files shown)


resources/ts/Components/IntegrationClientCredential.tsx
```diff
@@ -46,30 +45,18 @@ export const IntegrationClientCredentials = ({
...
-          {
-            label: "details.credentials.client_id",
-            value: client?.clientId,
-          },
-          {
-            label: "details.credentials.client_secret",
-            value: client?.clientSecret,
-          },
-        ];
+  const clientToShow = keycloakClient && !isLive ? keycloakClient : client;
```
simon-debruijn, line 49: It looks correct, but I think we better verify once more in the publiq platform frontend when to show which keys.

@grubolsch You once said the general `KEYCLOAK_ENABLED` variable was supposed to no longer be used?
Only the `KEYCLOAK_TEST_CLIENT_ENABLED`?

How should the frontend behave then on platform-test and on platform (prod)?
Should we change the frontend code to only use `KEYCLOAK_TEST_CLIENT_ENABLED`?
And once production is changed to keycloak with or without `KEYCLOAK_TEST_CLIENT_ENABLED` things should work?

grubolsch: yes, KEYCLOAK_ENABLED should be refactored away, we now use a combination of KEYCLOAK_CREATION_ENABLED, KEYCLOAK_TEST_CLIENT_ENABLED.

I am wondering if we cannot make the logic easier?
Just say, "if KEYCLOAK_CREATION_ENABLED is true AND there is a keycloak client, show that one. Otherwise, show the auth0 client".
Then you don't even need KEYCLOAK_TEST_CLIENT_ENABLED.

simon-debruijn: I think your suggestion checks out. If possible I would do it that way

Anahkiasen: Okay will switch the condition to that

LucWollants (approved):
Looks good. I think we can create a follow-up ticket to remove the now obsolete flag `KEYCLOAK_TEST_CLIENT_ENABLED`

PR #1479, opened 2024-10-23
PPF-452 Make key visibility more clear for admins

JonasVHG:
### Changed

- `app/Domain/Integrations/Integration`: Add `isKeyVisibleForEnvironment()`
- `app/Nova/Resources/Auth0Client`: Add column with `Visible for integrator` in adminPanel.
- `app/Nova/Resources/KeycloakClient`: Add column with `Visible for integrator` in adminPanel.
- `app/Nova/Resources/UiTiDv1`: Add column with `Visible for integrator` in adminPanel.

### Fixed

- `KeyVisibility` is now more clear for Admins.

---
Ticket: https://jira.publiq.be/browse/PPF-452

### DIFF
--- a/app/Domain/Integrations/Integration.php
+++ b/app/Domain/Integrations/Integration.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Domain\Integrations;
 
 use App\Auth0\Auth0Client;
+use App\Auth0\Auth0Tenant;
 use App\Domain\Contacts\Contact;
 use App\Domain\Coupons\Coupon;
 use App\Domain\KeyVisibilityUpgrades\KeyVisibilityUpgrade;
@@ -12,6 +13,8 @@ use App\Domain\Organizations\Organization;
 use App\Domain\Subscriptions\Subscription;
 use App\Keycloak\Client as KeycloakClient;
 use App\UiTiDv1\UiTiDv1Consumer;
+use App\UiTiDv1\UiTiDv1Environment;
+use Laravel\Prompts\Key;
 use Ramsey\Uuid\UuidInterface;
 
 final class Integration
@@ -228,6 +231,12 @@ final class Integration
             $this->urls,
             fn (IntegrationUrl $url) => $url->type->value === $type->value && $url->environment->value === $environment->value
         ));
+    }
+
+    public function isKeyVisibleForEnvironment(UiTiDv1Environment|Auth0Tenant|Environment $environment): bool
+    {
+        return $environment->value !== 'acc' &&
+            $this->status !== IntegrationStatus::Deleted;
     }
 
     public function toArray(): array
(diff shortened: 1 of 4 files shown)


app/Nova/Resources/Auth0Client.php
```diff
@@ -67,6 +71,19 @@ public function fields(NovaRequest $request): array
                     Auth0Tenant::Testing->value => Auth0Tenant::Testing->name,
                     Auth0Tenant::Production->value => Auth0Tenant::Production->name,
                 ]),
+            Text::make('Visible for integrator', static function (Auth0ClientModel $model) {
+                $auth0Client = $model->toDomain();
+                /** @var Integration $integration */
+                $integration = App::get(IntegrationRepository::class)->getById($auth0Client->integrationId);
+                $isVisible = $auth0Client->tenant !== Auth0Tenant::Acceptance &&
```
LucWollants, line 78: I think you could add an extra method `Integration::isKeyVisibleForEnvironment`. It requires only a conversion from a specific environment to the environment inside the Integration namespace. But all those enums have the same value.

JonasVHG: Wouldn't you also need to take the `keyVisibility` into account?

JonasVHG: But maybe I can get the visibilty from environment.

LucWollants: But the visibility is taken from the integration I think, see `$integration->getKeyVisibility()`

JonasVHG: I have implemented it, but I do not think it's ideal.

app/Domain/Integrations/Integration.php
```diff
@@ -230,6 +232,12 @@ public function urlsForTypeAndEnvironment(IntegrationUrlType $type, Environment
         ));
     }
 
+    public function isKeyVisibleForEnvironment(UiTiDv1Environment|Auth0Tenant|Environment $environment): bool
```
LucWollants, line 235: I think you can even do this

```
    public function isKeyVisibleForEnvironment(UiTiDv1Environment|Auth0Tenant|Environment $environment): bool
    {
        $keyVisibility = KeyVisibility::v1;
        if ($environment instanceof UiTiDv1Environment) {
            $keyVisibility = KeyVisibility::v2;
        }

        return $environment->value !== 'acc' &&
            $this->status !== IntegrationStatus::Deleted;
            $this->status !== IntegrationStatus::Deleted &&
            $this->getKeyVisibility() !== $keyVisibility;
    }
```

LucWollants: Or even work with a ternary operator.

LucWollants (commented):
Looks good, thanks for the extra refactor to a reusable method.

LucWollants (approved):

PR #1554, opened 2024-11-27
PPF-648 cleanup auth0 domain

JonasVHG:
### Changed

- `.env.ci`: Enable `Keycloak` instead of `auth0`
- `README.md`: Place info about `auth0` with `Keycloak`
- `config/app`: Unload `Auth0ServiceProvider`
- `AuthServiceProvider`: Always use `Keycloak` instead of looking at toggle & remove `getLoginParameters()`
- `LogoutController`: Always use `Keycloak` instead of looking at toggle.
- `AuthServiceProvider`: Always use `Keycloak` instead of looking at toggle.
- `Domain\Integration`: Remove property `$auth0Clients` & associated functions.
- `IntegrationModel`: Remove property `$auth0Clients` & associated functions.
- `tests/Keycloak/CachedKeycloakClientStatusTest`: Remove unused `CreatesMockAuth0ClusterSDK`
- `tests/Nova/ActionGuards/Keycloak/BlockKeycloakClientGuardTest`: Remove unused `CreatesMockAuth0ClusterSDK`
- `tests/Nova/ActionGuards/Keycloak/UnblockKeycloakClientGuardTest`: Remove unused `CreatesMockAuth0ClusterSDK`

### Removed

- `Console/Commands/Migrations/*`: No longer needed & depend on `auth0`
- `Console/Commands/ReadCsvFile`: No longer used
- `Console/Commands/UiTPAS/*`: No longer needed & depend on `auth0`
- `app/Auth0/*`
- `tests/Auth0/*`
- `tests/Console/Commands/Migrations/*`

---
Ticket: https://jira.publiq.be/browse/PPF-648


README.md
```diff
@@ -75,7 +75,6 @@ This requires:
 ## Integrations
 
 The following integrations are provided:
-- Auth0, can be disabled with `AUTH0_CLIENT_CREATION_ENABLED`
```
LucWollants, line 78: I think this can be replaced with Keycloak, it should also have a flag to control the creation of keys. For example when doing something in batch locally or on ACC and you don't want to pollute the external Keycloak database.

app/Providers/AuthServiceProvider.php
```diff
@@ -81,12 +74,7 @@ static function (): Auth0 {
     private function getLoginParameters(): array
     {
         $auth0LoginParameters = [];
-        if (config(KeycloakConfig::KEYCLOAK_LOGIN_ENABLED)) {
-            parse_str(config(KeycloakConfig::KEYCLOAK_LOGIN_PARAMETERS), $auth0LoginParameters);
-            return $auth0LoginParameters;
-        }
-
-        parse_str(config('auth0.login_parameters'), $auth0LoginParameters);
+        parse_str(config(KeycloakConfig::KEYCLOAK_LOGIN_PARAMETERS), $auth0LoginParameters);
```
LucWollants, line 77: Maybe we no longer need a method?

JonasVHG: I think not, you need the parameters as an array not a string, & config() returns it as a string?

LucWollants: I should have been more specific, I mean the method `getLoginParameters` ?

JonasVHG: Ah, now I get it.

PR #1594, opened 2024-12-17
Force a Nova publish

LucWollants:
### Changed
- Force a Nova publish on artifact build


lib/tasks/build.rake
```diff
@@ -7,6 +7,7 @@ task :build do |task|
   system('cp .env.ci .env') or exit 1
   system('php artisan package:discover') or exit 1
   system('php artisan horizon:install') or exit 1
+  system('php artisan nova:publish') or exit 1
```
willaerk, line 10: Is this the right place for this? Or do we need to add this to the deployment (in puppet)?

LucWollants: You are right, this should move to the deployment step.

PR #1632, opened 2025-01-17
Tweak condition showing "No credentials" alert to account for V2

Anahkiasen:
### Fixed
- Tweak condition showing "No credentials" alert to account for V2

---

Since the condition below uses that to pick which clients to render, and the condition for the alert only checked legacy clients, this seems to be the correct fix but it could be the things are more complicated than that? For example not sure if I should also check `keyVisibilityUpgrade`

Ticket: https://jira.publiq.be/browse/PPF-640


LucWollants (approved):
This will make the check indeed more reliable, but the main issue with this integration https://platform.publiq.be/admin/resources/integrations/e0f53aa5-c59e-405d-9948-138b37a8c62f is the missing v1 keys.

To get some feedback about failed key creation this ticket was created https://jira.publiq.be/browse/PPF-659

simon-debruijn (approved):

PR #1644, opened 2025-01-20
Don't just log but also capture Keycloak exceptions during client creation

Anahkiasen:
### Fixed
- Don't just log but also capture Keycloak exceptions during client creation

---

We seem to be throwing the Exception with the entire response from the API, so I don't think there is that much additional information to log that could help with the error in the screenshot since it seems Keycloak really is just returning "401 Unauthorized" without elaboration. 

So this change should at least prevent Keycloak exceptions from not reaching Sentry, as currently they are caught and logged but not released back for capture by Sentry? This will hopefully provide us with a lot more metadata to understand what goes wrong.

Ticket: https://jira.publiq.be/browse/PPF-659

### DIFF
--- a/app/Keycloak/Jobs/BlockClientHandler.php
+++ b/app/Keycloak/Jobs/BlockClientHandler.php
@@ -32,6 +32,7 @@ final readonly class BlockClientHandler implements ShouldQueue
                 [
                     'domain' => 'keycloak',
                     'id' => $event->id,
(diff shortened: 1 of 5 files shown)


LucWollants:
@Anahkiasen I guess we need to do this for all jobs because we don't report anything.

But I do wonder why the `error` method does not reach Sentry, because the default stack has Sentry included.

PR #1733, opened 2025-03-25
PPF-672 - Fix create uitpas integration

simon-debruijn:
### Fixed
- Fix create uitpas integration
- [Fix urls to create uitpas integration](https://github.com/cultuurnet/publiq-platform/pull/1733/commits/d8ad680341aff034e517e6020350c8052d9f6b3c)

### Changed
- [Always set and validate type on the backend](https://github.com/cultuurnet/publiq-platform/pull/1733/commits/0b0b2a8823b8f3345c5dce0289c513c52ab5c4ea)

---
Ticket: https://jira.uitdatabank.be/browse/PPF-672

### DIFF
--- a/app/Domain/Integrations/Controllers/IntegrationController.php
+++ b/app/Domain/Integrations/Controllers/IntegrationController.php
@@ -106,7 +106,9 @@ final class IntegrationController extends Controller
 
     public function create(Request $request): RedirectResponse|Response
     {
-        $integrationType = $request->query('type');
+        $integrationType = IntegrationType::tryFrom(
+            $request->query('type', '')
+        );
 
         if ($integrationType === null) {
             return Redirect::route(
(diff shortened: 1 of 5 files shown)


app/Domain/Integrations/Controllers/IntegrationController.php
```diff
@@ -104,8 +104,23 @@ public function index(Request $request): Response
         ]);
     }
 
-    public function create(): Response
+    public function create(Request $request): RedirectResponse|Response
     {
+        $type = $request->query('type', '');
+
+        $integrationType = IntegrationType::tryFrom(
+            is_array($type) ? $type[0] : $type
```
LucWollants, line 112: I wonder why this can be an array or a string?

simon-debruijn: e.g.

`?type=entry -> string`

or

`?type=entry&type=search -> string[]`

LucWollants: Yes, but it's a query parameter so some frontend code sets it as a string and other parts set it as an array?

simon-debruijn: A user can manually set a url. It is quite common in frontend frameworks aswel that you handle both cases

LucWollants (approved):

PR #1755, opened 2025-04-11
Do not allow everybody on admin panel on dev

JonasVHG:
### Changed

 - `NovaServiceProvider`: override `authorization()` to not allow everybody on localhost.

### Fixed

- The `local`-environment mimicks more closely the `server`-environments

---


grubolsch (approved):
Like the change, very pro active!

LucWollants:
> Like the change, very pro active!

Indeed, let's just try this. Maybe with a quick check of how the acceptance tests handle this?

PR #1831, opened 2025-06-04
PPF-664 Update permissions uitpas

grubolsch:
### Added
- As a user, when creating an UiTPAS integration, the test UDB3 organizers gets activated

### Background information
The introduction of UitpasHttpClient follows the same pattern as the existing Keycloak integration, where a dedicated HTTP client handles all API interactions. The key difference lies in how bearer tokens are obtained: Keycloak tokens are fetched from the master realm, while UiTPAS requires tokens from the UiTID realm. As a result, the UiTPAS implementation needs more configuration—specifically, it must support both the testing and production environments simultaneously.

I am currently listing to both the IntegrationCreated and IntegrationUpdated event, to catch instances where the type of an integration changes: this might be deemed overkill.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-664


LucWollants (commented):
Looks very clean and readable! Nice work!

As discussed through Slack a dedicated `UiTPASHttpClient` is not needed inside the `Keycloak` namespace, that can even further limit the amount of changes required.

grubolsch:
All feedback has been processed

app/UiTPAS/UiTPASApiInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace App\UiTPAS;
+
+use App\Keycloak\Realm;
+
+interface UiTPASApiInterface
+{
+    public function addPermissions(Realm $realm, string $organizerId, string $clientId): void;
```
LucWollants, line 11: I don't think `Realm` is needed in this context. I think the Keycloak Realm concept now leaks into a simple UiTPAS API call.

app/UiTPAS/Listeners/GiveUitpasPermissionsToTestOrganizer.php
```diff
@@ -0,0 +1,44 @@
...
+use App\Domain\Integrations\Events\IntegrationUpdated;
+use App\Domain\Integrations\IntegrationType;
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Keycloak\Realm;
+use App\UiTPAS\UiTPASApiInterface;
+use App\UiTPAS\UiTPASConfig;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+final class GiveUitpasPermissionsToTestOrganizer implements ShouldQueue
```
LucWollants, line 18: Does this class name really describe what happens? Is the keycloak client ID of the integration not added to the test organizer?

app/UiTPAS/UiTPASApiInterface.php
LucWollants: I think the concrete `UiTPASApi` needs to be injected with the `Realm` but not this interface.

app/UiTPAS/Listeners/GiveUitpasPermissionsToTestOrganizer.php
LucWollants: It think it is `AddUiTPASPermissionsToOrganizerForIntegration`

app/Keycloak/KeycloakConfig.php
```diff
@@ -6,11 +6,17 @@
...
     public const KEYCLOAK_CREATION_ENABLED = 'keycloak.creationEnabled';
     public const KEYCLOAK_DOMAIN = 'keycloak.login.domain';
     public const KEYCLOAK_CLIENT_ID = 'keycloak.login.clientId';
+    public const KEYCLOAK_CLIENT_SECRET = 'keycloak.login.clientSecret';
     public const KEYCLOAK_REALM_NAME = 'keycloak.login.realmName';
     public const KEYCLOAK_LOGIN_PARAMETERS = 'keycloak.login.parameters';
     public const KEYCLOAK_ENFORCE_2FA_FOR_ADMINS = 'keycloak.login.enforce2FAForAdmins';
+
+    // We need the Test env separately, because the UiTPAS integrations add new clients to a fixed test organizer
+    public const TEST_INTERNAL_NAME = 'keycloak.uitid_realms.test.internalName';
```
LucWollants, line 18: Not clear this is for UiTiD

app/UiTPAS/Listeners/AddUiTPASPermissionsToOrganizerForIntegration.php
```diff
@@ -0,0 +1,45 @@
...
+
+        if ($integration->type !== IntegrationType::UiTPAS) {
+            return;
+        }
+
+        $keycloakClient = $integration->getKeycloakClientByEnv(Environment::Testing);
+
+        $this->uitpasApi->addPermissions(
+            $this->testContext,
+            (string)config(UiTPASConfig::TEST_ORGANISATION->value),
```
LucWollants, line 41: If we always give hardcoded access to the organizer for the test environment, why do we even need the config for the production environment?

app/Keycloak/KeycloakConfig.php
grubolsch: Renamed the constants

app/Api/TokenStrategy/ClientCredentials.php
```diff
@@ -62,9 +69,21 @@ public function fetchToken(KeycloakHttpClient $client, Realm $realm): string
...
+        $this->logger->info('Fetched token for ' . $context->clientId . ', token starts with ' . substr($token, 0, 6));
+        $this->accessToken[$key] = $token;
+
+        return $token;
+    }
+
+    /**
+     * @throws GuzzleException
+     */
+    public function sendWithoutBearer(RequestInterface $request, ClientCredentialsContext $context): ResponseInterface
```
LucWollants, line 82: This can be a private method, not sure if we even need this method.

app/UiTPAS/ClientCredentialsContextFactory.php
```diff
@@ -0,0 +1,34 @@
...
+            config(KeycloakConfig::UITID_TEST_BASE_URL),
+            config(KeycloakConfig::UITID_CLIENT_ID),
+            config(KeycloakConfig::UITID_CLIENT_SECRET),
+            config(KeycloakConfig::UITID_TEST_INTERNAL_NAME),
+        );
+    }
+
+    public static function getUitIdProdContext(): ClientCredentialsContext
+    {
+        return new ClientCredentialsContext(
```
LucWollants, line 26: This doesn't seem to be used, and I also wonder why the test context starts with `UITID` and the production context starts with `KEYCLOAK`. For example, the ID and Secret are just UiTiD-related values and not directly related to `KEYCLOAK`

grubolsch: It will be needed to fetch permissions in the next pr, from the production server.
The reason I use different constants, is because to determine what the "production" server is, I will look at the config that is already in use for the login-functionality. This means that the "production server" is flexibel, while the "test" server is hardcoded, this will lead to a config like this:

**Acceptance server / local docker:**
Production server is actually the **acceptance** env
Test server is the test server

**Test server:**
Production server is actually the **test** env
Test server is the test server

**Production server**
Production server is actually the **production** env
Test server is the test server

I used to have a seperate config for this, but I thought you gave me feedback to remove it (not needed), but maybe I misunderstood.

config/keycloak.php
```diff
@@ -57,4 +58,19 @@
             ],
         ],
     ],
+    'uitid_realms' => [
+        // These are uitid realm clients (can be used for calls to other Publiq APIs like Uitpas
+        'test' => [
+            'internalName' => env('KEYCLOAK_TEST_UITID_REALM_NAME', ''),
+            'base_url' => env('KEYCLOAK_TEST_UITID_BASE_URL', ''),
```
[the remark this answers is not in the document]
grubolsch: Not sure if I agree.
It has nothing to do with Uitpas, it is generic UiTID config.
If you check the [appconfig PR](https://github.com/cultuurnet/appconfig/pull/1084 ) you can see none of the values are related to Uitpas.

Of course right now we only use it in the context of the uitpas integration, but that might not forever be the case.

app/UiTPAS/ClientCredentialsContextFactory.php
LucWollants: No, we should do this like we do for Keycloak.

app/UiTPAS/UiTPASConfig.php
```diff
@@ -0,0 +1,23 @@
...
+enum UiTPASConfig: string
+{
+    case REALM_NAME = 'uitid';
+
+    case AUTOMATIC_PERMISSIONS_ENABLED = 'uitpas.automatic_permissions_enabled';
+    case TEST_ORGANISATION = 'uitpas.test.organisation';
+    case TEST_API_ENDPOINT = 'uitpas.test.api_endpoint';
+    case PROD_API_ENDPOINT = 'uitpas.prod.api_endpoint';
+
+    case UITPAS_TEST_OAUTH_TOKEN_URL = 'uitpas.test.oath_token_url';
```
LucWollants, line 16: Do we need the `UITPAS` prefix?

tests/Uitpas/Listeners/GiveUitpasPermissionsToTestOrganizerTest.php
```diff
@@ -0,0 +1,131 @@
...
+            Uuid::uuid4(),
+            IntegrationType::UiTPAS,
+            'My uitpast test',
+            'Lorum ipsum',
+            Uuid::uuid4(),
+            IntegrationStatus::Draft,
+            IntegrationPartnerStatus::THIRD_PARTY,
+        ))->withKeycloakClients(
+            ... [
+                new Client(Uuid::uuid4(), Uuid::uuid4(), '5f263a50-9474-4690-a962-6935d6f9a3f2', 'client-test', Environment::Testing),
```
LucWollants, line 69: I guess you can use `$clientIdTest ` here?

app/UiTPAS/UiTPASServiceProvider.php
```diff
@@ -0,0 +1,55 @@
...
+            return;
+        }
+
+        $this->bootstrapEventHandling();
+    }
+
+    private function bootstrapEventHandling(): void
+    {
+        Event::listen(IntegrationCreated::class, [AddUiTPASPermissionsToOrganizerForIntegration::class, 'handle']);
+        Event::listen(IntegrationUpdated::class, [AddUiTPASPermissionsToOrganizerForIntegration::class, 'handle']);
```
LucWollants, line 53: Do we need to handle the `IntegrationUpdated` event?

LucWollants (approved):
Thanks for all the extra tweaks, nice work! I only had some minor remarks, so I already approved.

PR #1844, opened 2025-06-11
PPF-667 Slack notification for udb organizer

grubolsch:
### Added
- Added new Event called UdbOrganizerAdded, when triggered it sends a slack message to a new channel
- Has configurable uri for uitpas website and udb3 website for slack message
- Bonus feature: only allow adding of organisations when this org exists in UDB3
- Bonus feature: throw an user friendly error when trying to add an org that already was added

---
Ticket: https://jira.uitdatabank.be/browse/PPF-667

### DIFF
--- a/app/UiTPAS/UiTPASServiceProvider.php
+++ b/app/UiTPAS/UiTPASServiceProvider.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace App\UiTPAS;
 
 use App\Api\TokenStrategy\ClientCredentials;
-use App\Domain\Integrations\Events\IntegrationCreated;
 use App\Domain\Integrations\Events\UdbOrganizerCreated;
 use App\Domain\Integrations\GetIntegrationOrganizersWithTestOrganizer;
 use App\Domain\Integrations\Repositories\IntegrationRepository;
--- a/app/Notifications/NotificationsProvider.php
+++ b/app/Notifications/NotificationsProvider.php
@@ -10,7 +10,7 @@ use App\Domain\Subscriptions\Repositories\SubscriptionRepository;
 use App\Notifications\Listeners\NotifyIntegrationChanged;
 use App\Notifications\Slack\SlackMessageBuilder;
 use App\Notifications\Slack\SlackNotifier;
-use App\Search\FetchNameForUdb3Organizer;
+use App\Search\UdbOrganizerNameResolver;
 use App\Search\Sapi3\SearchService;
 use App\UiTPAS\UiTPASConfig;
 use Illuminate\Support\Facades\Event;
@@ -35,7 +35,7 @@ final class NotificationsProvider extends ServiceProvider
             fn () =>
             new SlackMessageBuilder(
                 $this->app->get(SubscriptionRepository::class),
-                $this->app->get(FetchNameForUdb3Organizer::class),
+                $this->app->get(UdbOrganizerNameResolver::class),
                 $this->app->get(SearchService::class),
                 config(UiTPASConfig::CLIENT_PERMISSIONS_URI->value),
                 config(UiTPASConfig::UDB_BASE_URI->value),
--- a/app/Notifications/Slack/SlackMessageBuilder.php
+++ b/app/Notifications/Slack/SlackMessageBuilder.php
@@ -10,14 +10,14 @@ use App\Domain\Integrations\IntegrationStatus;
 use App\Domain\Integrations\UdbOrganizer;
 use App\Domain\Subscriptions\Repositories\SubscriptionRepository;
 use App\Notifications\MessageBuilder;
-use App\Search\FetchNameForUdb3Organizer;
+use App\Search\UdbOrganizerNameResolver;
 use App\Search\Sapi3\SearchService;
 
 final readonly class SlackMessageBuilder implements MessageBuilder
 {
     public function __construct(
         private SubscriptionRepository $subscriptionRepository,
-        private FetchNameForUdb3Organizer $fetchNameForUdb3Organizer,
+        private UdbOrganizerNameResolver $fetchNameForUdb3Organizer,
         private SearchService $searchService,
         private string $uitpasRootUri,
         private string $udbRootUri,
@@ -45,7 +45,7 @@ final readonly class SlackMessageBuilder implements MessageBuilder
     {
         $client = $integration->getKeycloakClientByEnv(Environment::Production);
 
-        $organizerName = $this->fetchNameForUdb3Organizer->fetchName($this->searchService->findUiTPASOrganizers($udbOrganizer->organizerId));
+        $organizerName = $this->fetchNameForUdb3Organizer->getName($this->searchService->findUiTPASOrganizers($udbOrganizer->organizerId));
 
         $message = '*:robot_face: :incoming_envelope: ' . $integration->name . ' - requested access to organisation ' . $organizerName . ' (' . $udbOrganizer->organizerId . ') *';
         $message .= PHP_EOL . PHP_EOL;
--- a/app/Nova/Resources/UdbOrganizer.php
+++ b/app/Nova/Resources/UdbOrganizer.php
@@ -6,7 +6,7 @@ namespace App\Nova\Resources;
 
 use App\Domain\Integrations\Models\UdbOrganizerModel;
 use App\Nova\Resource;
-use App\Search\FetchNameForUdb3Organizer;
+use App\Search\UdbOrganizerNameResolver;
 use App\Search\Sapi3\SearchService;
 use App\UiTPAS\UiTPASConfig;
 use Illuminate\Support\Facades\App;
@@ -50,8 +50,8 @@ final class UdbOrganizer extends Resource
                 ->readonly(),
 
             Text::make('Name', static function (UdbOrganizerModel $model) {
-                /** @var FetchNameForUdb3Organizer $fetchNameForUdb3Organizer */
-                $fetchNameForUdb3Organizer = App::get(FetchNameForUdb3Organizer::class);
+                /** @var UdbOrganizerNameResolver $udbOrganizerNameResolver */
+                $udbOrganizerNameResolver = App::get(UdbOrganizerNameResolver::class);
 
                 /** @var SearchService $searchService */
                 $searchService = App::get(SearchService::class);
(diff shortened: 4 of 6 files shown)


app/Nova/Actions/AddUdbOrganizer.php
```diff
@@ -36,13 +38,21 @@ public function handle(ActionFields $fields, Collection $integrations): ActionRe
...
+        try {
+            $this->organizerRepository->create(
+                new UdbOrganizer(
+                    $id,
+                    Uuid::fromString($integration->id),
+                    $organizationIdAsString
+                )
+            );
+        } catch (PDOException) {
+            return Action::danger('Organizer "' . $organizationIdAsString . '" was already added.');
```
LucWollants, line 52: There could be other reasons this fails with a `PDOException`

app/UiTPAS/Listeners/SendSlackMessageWhenOrganizerIsRequested.php
```diff
@@ -0,0 +1,52 @@
...
+use App\Domain\Integrations\Repositories\IntegrationRepository;
+use App\Domain\Integrations\Repositories\UdbOrganizerRepository;
+use App\Notifications\Notifier;
+use App\UiTPAS\Slack\UdbOrganizerMessageBuilder;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+final class SendSlackMessageWhenOrganizerIsRequested implements ShouldQueue
```
LucWollants, line 18: I would make it more general, like `NotifyIntegrationChanged` and change it to something like `NotifyUdbOrganizerAdded`

app/UiTPAS/Slack/UdbOrganizerMessageBuilder.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace App\UiTPAS\Slack;
+
+use App\Domain\Integrations\Environment;
+use App\Domain\Integrations\Integration;
+use App\Domain\Integrations\UdbOrganizer;
+
+final readonly class UdbOrganizerMessageBuilder
```
LucWollants, line 11: There seems to be a lot of common code with the existing `SlackMessageBuilder`

tests/Domain/Integrations/Repositories/EloquentUdbOrganizerRepositoryTest.php
```diff
@@ -79,4 +79,11 @@ public function testDelete(): void
             'organizer_id' => $this->organizer1->organizerId,
         ]);
     }
+
+    public function test_it_can_get_an_udb_organizer_by_id(): void
```
LucWollants, line 83: Not related to this pull request, but I think the `testDelete` method doesn't create the `$organizer1` inside the database.

app/UiTPAS/Listeners/SendSlackMessageWhenOrganizerIsRequested.php
grubolsch: I will got for `NotifyIntegrationRequested`, to communicate this is part of an approval flow.

app/Domain/Integrations/Events/UdbOrganizerAdded.php
```diff
@@ -0,0 +1,17 @@
...
+namespace App\Domain\Integrations\Events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+use Ramsey\Uuid\UuidInterface;
+
+final readonly class UdbOrganizerAdded
+{
+    use Dispatchable;
+
+    public function __construct(public UuidInterface $id)
```
LucWollants, line 14: I wonder if this should not contain the integration ID and the organizer ID? Because an UDB3 organizer is added to an integration, so to define the event uniquely both IDs are required.

app/UiTPAS/Slack/UdbOrganizerMessageBuilder.php
grubolsch: Not really? I did look into reusing the existing `SlackMessageBuilder` but the main reason i decided not to use it is because the inferface requires an `Integration`, but in my case I start from an `UdbOrganizer`.
So just making another implementation was difficult.

I can however, merge both classes into 1 if you prefer, with a second method?
Don't really see the benefit, but that is an option.

tests/Domain/Integrations/Repositories/EloquentUdbOrganizerRepositoryTest.php
grubolsch: haha, good catch, will fix it.

app/Domain/Integrations/Events/UdbOrganizerAdded.php
grubolsch: Renamed to `UdbOrganizerCreated`

app/UiTPAS/UiTPASServiceProvider.php
```diff
@@ -50,6 +55,27 @@ public function register(): void
             );
         });
 
+        $this->app->singleton(UdbOrganizerMessageBuilder::class, function () {
```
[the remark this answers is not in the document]
LucWollants: I mean you can directly inject the `UdbOrganizerMessageBuilder` inside the constructor of `SendSlackMessageWhenOrganizerIsRequested` on line [75](https://github.com/cultuurnet/publiq-platform/pull/1844/files/7e5641c20353bcaceb4ae459486d5fd3a663fca7#diff-208fec38076ede4b286fa5c4ee114767e3d4802e786a7a454f34d388baac6dd4R75).

app/UiTPAS/Slack/UdbOrganizerMessageBuilder.php
LucWollants: Now I also see this class does not derive from `MessageBuilder`.

But I still think there is a lot in common, for example:
- Message title
- Type
- Status
- Subscription
- Website
- Open in publiq-platform

The only extra things are open in UiTPAS and open in UDB3

LucWollants: There is also nothing wrong with adding an extra method `toMessageWithOrganizer` and implementing that inside the existing `SlackMessageBuilder`. That is also one of the reasons this concrete `SlackMessageBuilder` is inside the `Notifications` namespace.

app/UiTPAS/Slack/UdbOrganizerMessageBuilder.php
```diff
@@ -0,0 +1,35 @@
...
+        $client = $integration->getKeycloakClientByEnv(Environment::Production);
+
+        $message = '*:robot_face: :incoming_envelope: ' . $integration->name . ' - requested access to organisation ' . $org->organizerId . '*';
+        $message .= PHP_EOL . PHP_EOL;
+        $message .= PHP_EOL . '• *Status:* _' . $integration->status->value . '_';
+        $message .= PHP_EOL . '• *Description:* _' . $integration->description . '_';
+        $message .= PHP_EOL . '• *Website:* _' . ($integration->website() ? $integration->website()->value : 'N/A') . '_';
+        $message .= PHP_EOL;
+        $message .= PHP_EOL . '• *Open in publiq-platform:* ' . $this->appUrl . '/admin/resources/integrations/' . $integration->id->toString();
+        //@todo Not sure if we want to localize this URL on the env?
```
LucWollants, line 30: If we want to have some easy and realistic tests for ACC and TEST we need to use the environment.

app/Nova/Actions/AddUdbOrganizer.php
```diff
@@ -67,4 +74,10 @@ public function fields(NovaRequest $request): array
                 ),
         ];
     }
+
+    private function doesOrganisationExistInUdb3(string $organizerId): bool
```
LucWollants, line 78: Maybe keep it consistent and use `organizerExistInUdb`

app/Search/Sapi3/Sapi3SearchService.php
```diff
@@ -34,6 +34,7 @@ public function findUiTPASOrganizers(string ...$ids): PagedCollection
             return new PagedCollection();
         }
 
+        //@todo I wonder if there is a bug here, because the function has a constraint on labels:uitpas, but this one does not?
```
LucWollants, line 37: No, because this uses the provided IDs instead of searching an UiTPAS organizer on name.

app/UiTPAS/FetchNameForUdb3Organizer.php
```diff
@@ -0,0 +1,33 @@
...
+namespace App\UiTPAS;
+
+use App\Search\Sapi3\SearchService;
+use CultuurNet\SearchV3\ValueObjects\Organizer;
+
+/*
+ * If performance starts to be an issue, a cache would be good here.
+ * */
+
+final readonly class FetchNameForUdb3Organizer
```
LucWollants, line 14: This feels like overkill compared to this https://github.com/cultuurnet/publiq-platform/blob/8ed64587e4d3db6ff2d92a57e3888b3dcc83b066/app/Domain/Integrations/GetIntegrationOrganizersWithTestOrganizer.php#L39

LucWollants: If you want to make it reusable I would just make it part of `SearchService`

app/UiTPAS/FetchNameForUdb3Organizer.php
```diff
@@ -0,0 +1,33 @@
...
+        $result = $this->searchService->findUiTPASOrganizers($organizerId);
+
+        if ($result->getTotalItems() < 1) {
+            return 'Niet teruggevonden in UDB3';
+        }
+
+        /** @var Organizer $organizer */
+        $organizer = $result->getMember()?->getItems()[0];
+
+        return $organizer->getName()?->getValueForLanguage('nl') ?? $organizer->getName()?->getValueForLanguage('fr') ?? '';
```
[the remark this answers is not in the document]
grubolsch: Thanks for the tip!
MainLanguage can be null, so I did provide a fallback to NL.

app/UiTPAS/FetchNameForUdb3Organizer.php
LucWollants: I don't agree that it does 1 thing, it does 2 things. It gets the organizer and it gets the name. If you follow the SOLID principle, you would create a class that works on the Organizer collection and fetches the name of an already retrieved organizer.

grubolsch: oké, I moved the dependency for the SearchService outside the class, and passed a collection.
Does this work for you?

LucWollants: I would now also move this to the `Search` namespace and think about the name of the class and method. Something like `UdbOrganizerNameResolver::getName`

PR #1858, opened 2025-06-17
PPF-385 Change when the permissions for UiTPAS are created, do it after keycloak

grubolsch:
### Changed
- Changes when UiTPAS permissions are set for an integration.
Instead of doing it after `IntegrationCreated`, the permissions are now added only after the corresponding Keycloak client has been successfully created. This guarantees that Keycloak is fully ready before UiTPAS calls are made, because on the servers these calls are processed async. 

---
Ticket: https://jira.uitdatabank.be/browse/PPF-385

### DIFF
--- a/app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
+++ b/app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
@@ -7,7 +7,6 @@ namespace App\Keycloak\Repositories;
 use App\Domain\Integrations\Environment;
 use App\Domain\Integrations\Environments;
 use App\Keycloak\Client;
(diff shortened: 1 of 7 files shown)


app/Keycloak/Repositories/EloquentKeycloakClientRepository.php
```diff
@@ -38,6 +39,8 @@ public function create(Client ...$clients): void
                             'realm' => $this->realms->getRealmByEnvironment($client->environment)->publicName,
                         ]
                     );
+
+                ClientCreated::dispatch($client->id);
```
LucWollants, line 43: If you do this for every single Client in this loop, then I would move it to the model class.

PR #1859, opened 2025-06-18
PPF-670 Approve reject new udb3 organizer request

grubolsch:
### Added
- Added a status field for UdbOrganizer
- Add 2 actions for the organizer, to approve or reject an request.
- Add UiTPAS permissions to accepted organizers

### Changed
- Can no longer delete UdbOrganizer in the ui, need to be rejected for deletion.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-670


app/Domain/Integrations/UdbOrganizer.php
```diff
@@ -11,7 +11,18 @@
     public function __construct(
         public UuidInterface $id,
         public UuidInterface $integrationId,
-        public string $organizerId
+        public string $organizerId,
+        public UdbOrganizerStatus $status = UdbOrganizerStatus::Pending,
```
LucWollants, line 15: This shouldn't be part of the default constructor, see `Integration::withKeyVisibility`

app/Domain/Integrations/Repositories/EloquentUdbOrganizerRepository.php
```diff
@@ -30,6 +31,21 @@ public function createInBulk(UdbOrganizers $organizers): void
         });
     }
 
+    public function save(UdbOrganizer $organizer): void
+    {
+        UdbOrganizerModel::query()->updateOrCreate(
```
LucWollants, line 36: Do we need an `updateOrCreate`? Can't we just use an `updateStatus`?

app/Nova/Actions/ActivateUdbOrganizer.php
```diff
@@ -0,0 +1,41 @@
...
+    public function __construct(
+        private readonly Dispatcher $dispatcher,
+        private readonly ActivateUiTPASClientHandler $listener,
+    ) {
+    }
+
+    public function handle(ActionFields $fields, Collection $udbOrganizers): void
+    {
+        foreach ($udbOrganizers as $udbOrganizer) {
+            if (!$udbOrganizer instanceof UdbOrganizerModel) {
```
LucWollants, line 34: This should never happen. If this were to happen, then we need to fix it.

app/UiTPAS/Jobs/ActivateUiTPASClientHandler.php
```diff
@@ -0,0 +1,42 @@
...
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+final class ActivateUiTPASClientHandler implements ShouldQueue
+{
+    use Queueable;
+
+    public function __construct(
+        private readonly UdbOrganizerRepository $udbOrganizerRepository,
+        private readonly IntegrationRepository $integrationRepository,
+        private readonly UiTPASApiInterface $api,
```
LucWollants, line 22: I would make the name of the property more explicit

config/uitpas.php
```diff
@@ -15,7 +15,7 @@
     ],
     'prod' => [
         'api_endpoint' => env('UITPAS_PROD_ENDPOINT', ''),
-        'oath_token_url' => env('UITPAS_PROD_OATH_TOKEN_URL', ''),
+        'oath_token_url' => env('UITPAS_PROD_OAUTH_TOKEN_URL', ''),
```
LucWollants, line 18: There is still a typo here `oath_token_url`

app/Domain/Integrations/UdbOrganizer.php
grubolsch: I am happy to change it but can you explain a little bit why?
I considered your approach, but I went with a default value because this lowers the amount of files/code I have to change, because this object is created in 15 different places.

app/Nova/Actions/ActivateUdbOrganizer.php
grubolsch: I agree this should never happen, and probably never will.
The reason I write code like this is twofold:
- Just pure defensive programming
- This helps the IDE understand the type of `$udbOrganizer`,  without using docblocks. 

I would prefer to keep these types of statements.

LucWollants: The side effect of this defensive style is that errors go unnoticed. To help the IDE understand the type of the Collection, you can also typehint that.

app/Nova/Actions/RejectUdbOrganizer.php
```diff
@@ -0,0 +1,34 @@
...
+    public function __construct(
+        private readonly UdbOrganizerRepository $udbOrganizerRepository
+    ) {
+    }
+
+    public function handle(ActionFields $fields, Collection $udbOrganizers): void
+    {
+        /** @var UdbOrganizerModel $udbOrganizer */
+        foreach ($udbOrganizers as $udbOrganizer) {
+            $this->udbOrganizerRepository->delete($udbOrganizer->toDomain());
```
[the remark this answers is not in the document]
grubolsch: I discussed this with Jean-Marie, and we opted to just delete it, mainly because currently you cannot re-request an organisation because of a unique key constraint on `organisator_id` and `integration_id`.
You can also not give a reason for the rejection, so the requester would not get much info out of the rejection.
He will however in the future get an automatic email to inform him his request was denied.

In practice, rejections are rare, and when they occur Manon often contacts them anyway with an explanation, and advice how to re-apply.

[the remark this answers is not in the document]
LucWollants: Ok clear! Thanks for explaining this.

I wonder if we should not just keep the delete action, and a side effect of this delete will be the removal of the permissions? Keep in mind that a reject or delete will only remove the organizer from the publiq-platform database but won't remove any permissions. So the integrator can still use the organizer as is.

app/Nova/Actions/ActivateUdbOrganizer.php
grubolsch: There is no error in this case breaking the code? Because I handled it with the if.
How would you typehint this without a docblock?

LucWollants: I said that coding errors would go unnoticed, not that it would result in errors. If, for some reason, the provided collection does not contain elements of type `UdbOrganizerModel`, there won't be a crash. But the result would be that organizers don't get activated. You obscure the flow of the code.

LucWollants: Also, this is code deep inside a domain layer we fully control. This is not something a user has entered and needs extra validation. In fully typed languages with generics this would fail when calling the method, in PHP this will fail when the method is executing.

LucWollants: But fine if you want to keep it.

app/Nova/Actions/RejectUdbOrganizer.php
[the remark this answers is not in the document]
grubolsch: I personally really like the explicitness of the approval / reject flow, and would find it strange if there was only an approval action, without rejection.
That being said, I am not against also providing the delete action.
if I understand you correctly you then would want the extra feature of removing all permissions for accepted organizers? I am happy to do that, but maybe in an seperate pr/ticket because this is a new feature?

app/UiTPAS/Jobs/ActivateUiTPASClientHandler.php
grubolsch: class does not exist anymore, moved to ActivateUdbOrganizer, where I renamed the param

app/Domain/Integrations/UdbOrganizer.php
grubolsch: After slack discussion, we agreed on making it required.

app/UiTPAS/UiTPASServiceProvider.php
```diff
@@ -72,8 +73,8 @@ public function register(): void
             );
         });
 
-        $this->app->singleton(ActivateUiTPASClientHandler::class, function () {
-            return new ActivateUiTPASClientHandler(
+        $this->app->singleton(ActivateUdbOrganizer::class, function () {
+            return new ActivateUdbOrganizer(
```
LucWollants, line 76: I don't think we need to register Nova actions?

grubolsch: In this case we do to resolve ` ClientCredentialsContextFactory::getUitIdProdContext()`, there is also a test context.

app/Domain/Integrations/Repositories/EloquentUdbOrganizerRepository.php
```diff
@@ -30,6 +32,16 @@ public function createInBulk(UdbOrganizers $organizers): void
         });
     }
 
+    public function updateStatus(string $organizerId, UdbOrganizerStatus $newStatus): void
```
LucWollants, line 35: Should the `organizerId` be typed on `UuidInterface`?

app/UiTPAS/UiTPASServiceProvider.php
```diff
@@ -71,6 +72,15 @@ public function register(): void
...
+                $this->app->get(UdbOrganizerRepository::class),
+                $this->app->get(IntegrationRepository::class),
+                $this->app->get(UiTPASApiInterface::class),
+                ClientCredentialsContextFactory::getUitIdProdContext(),
+            );
+        });
+
         if (!config(UiTPASConfig::AUTOMATIC_PERMISSIONS_ENABLED->value)) {
```
LucWollants, line 84: This is a very weird approach to disable `bootstrapEventHandling`

grubolsch: I assume you want the if inside the function? I will move it

LucWollants (approved):
I already approved. There was only one minor remark left, see https://github.com/cultuurnet/publiq-platform/pull/1859#discussion_r2161900111

PR #1871, opened 2025-06-23
PPF-670 Improvements uitpas integration

grubolsch:
### Changed
- Changed default permissions based on feedback from Jean-Marie.

### Fixed
- I missed a detail in the implementation of updatePermissions(), the API endpoint does a full replace of all permissions, so you have to sent all current permissions are your lose all other organisations when multiple are requested for the same client.
- Has a unit test to prove new functionality

---
Ticket: https://jira.uitdatabank.be/browse/PPF-670

### DIFF
--- a/tests/UiTPAS/UitpasApiTest.php
+++ b/tests/UiTPAS/UitpasApiTest.php
@@ -135,6 +135,7 @@ final class UitpasApiTest extends TestCase
 
         $uitpasApi->addPermissions($this->context, new UdbUuid(self::ORG_ID), self::CLIENT_ID);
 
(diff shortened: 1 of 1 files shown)


app/UiTPAS/UiTPASApi.php
```diff
@@ -31,10 +32,17 @@ public function __construct(
 
     public function addPermissions(ClientCredentialsContext $context, UdbUuid $organizerId, string $clientId): bool
     {
+        $response = $this->sendWithBearer(
```
LucWollants, line 35: Why don't you use the existing `fetchPermissions` method?

PR #1899, opened 2025-07-15
PPF-692 Hide, but not remove CouponCode

JonasVHG:
### Fixed

---
Ticket: https://jira.publiq.be/browse/PPF-692

### DIFF
--- a/app/Nova/Actions/ActivateIntegration.php
+++ b/app/Nova/Actions/ActivateIntegration.php
@@ -36,13 +36,10 @@ final class ActivateIntegration extends Action
         $organizationIdAsString = $fields->get('organization');
         $organizationId = Uuid::fromString($organizationIdAsString);
 
-        /** @var string $couponCode */
-        $couponCode = $fields->get('coupon');
-
         $this->integrationRepository->activateWithOrganization(
             Uuid::fromString($integration->id),
             $organizationId,
-            $couponCode
+            null
         );
 
         return Action::message('Integration "' . $integration->name . '" activated.');
@@ -58,12 +55,6 @@ final class ActivateIntegration extends Action
                 ->rules(
                     'required',
                     'exists:organizations,id'
-                ),
-            Text::make('Coupon', 'coupon')
-                ->rules(
-                    'nullable',
-                    'string',
-                    'exists:coupons,code'
                 ),
         ];
     }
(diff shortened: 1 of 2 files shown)


app/Nova/Actions/ActivateIntegration.php
```diff
@@ -59,12 +55,6 @@ public function fields(NovaRequest $request): array
                     'required',
                     'exists:organizations,id'
                 ),
-            Text::make('Coupon', 'coupon')
```
grubolsch, line 62: Can't we put this behind a feature flag that for now we disable? Because we might want to reuse this in the future?

app/Nova/Actions/ActivateIntegration.php
```diff
@@ -36,13 +35,10 @@ public function handle(ActionFields $fields, Collection $integrations): ActionRe
...
         $organizationId = Uuid::fromString($organizationIdAsString);
 
-        /** @var string $couponCode */
-        $couponCode = $fields->get('coupon');
-
         $this->integrationRepository->activateWithOrganization(
             Uuid::fromString($integration->id),
             $organizationId,
-            $couponCode
+            null
```
grubolsch, line 41: Do we have to do this? I would not change this code. If the field is empty (because it is not included in the form, it should default to null anyway.

resources/ts/Pages/Integrations/New.tsx
```diff
@@ -83,10 +83,7 @@ const New = ({ subscriptions }: Props) => {
     });
   }
 
-  const isCouponFieldVisible =
-    (activeType === IntegrationType.SearchApi ||
-      activeType === IntegrationType.Widgets) &&
-    data.subscriptionId === basicPricingPlan?.id;
+  const isCouponFieldVisible = false;
```
grubolsch, line 86: Can we also hide this behind a feature flag instead?

grubolsch (changes requested):

grubolsch:
ik denk dat we dit op een andere manier kunnen aanpakken zodat we minder code moeten veranderen en sneller de feature kunnen inschakelen terug. Ik zou met een feature flag werken.

config/coupons.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+declare(strict_types=1);
+
+return [
+    'enabled' => env('COUPONS_ENABLED', true),
```
LucWollants, line 6: It feels a bit weird to create a separate file for this, I would add a `features` part to `app.php`

PR #1900, opened 2025-07-15
PPF-693 Cleanup permissions when revoking organizer

grubolsch:
### Added
- Added new option to an approved organizer: "revoke permissions". This will remove the organizer from the db and revoke his permissions in UiTPAS.

### Changed
- Admin added organisations are directly approved, and only recieve the final "approved" email.
- Made status field for organizer visually the same as other status fields in PPF. (color = green)

### Fixed
- UpdateStatus() organizer method was bugged, updated all data, without a where-statement.

---
Ticket: https://jira.uitdatabank.be/browse/PPF-693


JonasVHG (approved):
Looks good.

PR #1902, opened 2025-07-17
BUGFIX Improve SMTP DSN construction for mailer

grubolsch:
### Fixed

- Use username and password in DSN only if both are set
- Fallback to host and port if credentials are missing

### Background
The current code fails on acc because there is no username/password there, because mails are configured to be sent to mailtrap.

screenshot_2025-07-16_at_16 39 26


app/UiTPAS/UiTPASServiceProvider.php
```diff
@@ -122,4 +120,25 @@ private function bootstrapEventHandling(): void
         Event::listen(UdbOrganizerApproved::class, [SendMailForUdbOrganizer::class, 'handleUdbOrganizerApproved']);
         Event::listen(UdbOrganizerRejected::class, [SendMailForUdbOrganizer::class, 'handleUdbOrganizerRejected']);
     }
+
+    private function getSmtpDsn(): string
```
LucWollants, line 124: Can we not just add the full DSN to the config and load it from the `.env` file? So we add a `MAIL_DSN`, which is either the AWS setup or the mailhog setup. If we use that approach, we can even remove the username and password config.

This also makes me wonder if we need another pull request to change from mailhog to mailpit

PR #1936, opened 2025-08-04
Fetch labels uitpas from UDB + Cache

grubolsch:
### Changed
- Fetch labels uitpas from UDB 
- Caching the result

---


JonasVHG (approved):

LucWollants:
Is there no associated pull request to change the config?

LucWollants (changes requested):
I wonder if this approach is the best:
- I would expect a HTTP request which handles errors, timeouts and content type properly instead of a `file_get_contents`
- Injecting the cache directly is a bit weird, I would expect some kind of interface to get the UiTPAS labels which has a cached implementation
- I wonder how the unit tests still work without any changes? So that means that it was never tested that the UiTPAS labels are added?

grubolsch:
> I wonder if this approach is the best:
> 
> * I would expect a HTTP request which handles errors, timeouts and content type properly instead of a `file_get_contents`
> * Injecting the cache directly is a bit weird, I would expect some kind of interface to get the UiTPAS labels which has a cached implementation
> * I wonder how the unit tests still work without any changes? So that means that it was never tested that the UiTPAS labels are added?

All of these are great points, processed them all, there is now a decorator for the cache. Also much cleaner to test.
Also added a bunch of tests.

app/Search/SearchServiceProvider.php
```diff
@@ -6,25 +6,38 @@
...
-                        'headers' => [
-                            'X-Api-Key' => config('search.api_key'),
-                        ],
-                    ]),
+                    $client,
                     new Serializer()
+                ),
+                new CachedUiTPASLabelProvider(
+                    new HttpUiTPASLabelProvider(
+                        $client,
```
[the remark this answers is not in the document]
LucWollants: You also don't need an API key inside the header to get the public endpoint with UiTPAS labels

tests/Search/UiTPAS/HttpUiTPASLabelProviderTest.php
```diff
@@ -0,0 +1,71 @@
...
+        );
+
+        $labels = $provider->getLabels();
+
+        $this->assertSame([], $labels);
+    }
+
+    private function createStreamMock(string $contents): StreamInterface&MockObject
+    {
+        $stream = $this->getMockBuilder(StreamInterface::class)->getMock();
```
LucWollants, line 67: Does `$this->createMock(StreamInterface::class);` not work here?

tests/Search/Sapi3/Sapi3SearchServiceTest.php
```diff
@@ -0,0 +1,68 @@
...
+        $uuid1 = new UdbUuid(Uuid::uuid4()->toString());
+        $uuid2 = new UdbUuid(Uuid::uuid4()->toString());
+
+        $searchClient = $this->createMock(SearchClientInterface::class);
+        $searchClient->expects($this->once())
+            ->method('searchOrganizers')
+            ->with($this->callback(function (SearchQuery $result) use ($uuid1, $uuid2) {
+                $params = $result->toArray();
+
+                return str_contains($params['q'], 'id:"' . $uuid1->toString() . '"') &&
```
LucWollants, line 58: I wonder, is it not easier to construct the expected `SearchQuery` and compare it with the actual one?

PR #1937, opened 2025-08-04
PPF-700 SMTP UiTPAS mail improvements

grubolsch:
This pull request improves and extends the UiTPAS SMTP mails, adds a new mail, and adds automated Mailpit tests for it.

### Added
- Automated mailpit tests, but they don't run as a default (disabled in phpunit.xml)
- Added new email "Integration created", replacing the old Mailjet version (new copy).
- Introduced new event `UdbOrganizerRequested` that sends "activation requested" mails, instead of the "UdbOrganizerCreated" event for clarity
- Also listen to `IntegrationActivationRequested` to sent activation requested mail, but this variation bundles all requested orgs so user does not recieve multiple activation mails for one action.

### Changed
- Improved copy and added buttons to mail templates based on feedback
- Use directly DSN instead of building DSN config from seperate parts
- Improved/added many tests

---

### DIFF
--- a/app/Search/Sapi3/Sapi3SearchService.php
+++ b/app/Search/Sapi3/Sapi3SearchService.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Search\Sapi3;
 
 use App\Domain\UdbUuid;
+use App\Search\UiTPAS\UiTPASLabelProvider;
 use CultuurNet\SearchV3\Parameter\Query;
 use CultuurNet\SearchV3\SearchClientInterface;
 use CultuurNet\SearchV3\SearchQuery;
@@ -13,7 +14,8 @@ use CultuurNet\SearchV3\ValueObjects\PagedCollection;
 final readonly class Sapi3SearchService implements SearchService
 {
     public function __construct(
-        private SearchClientInterface $searchClient
+        private SearchClientInterface $searchClient,
+        private UiTPASLabelProvider $uiTPASLabelProvider,
     ) {
     }
 
@@ -43,9 +45,14 @@ final readonly class Sapi3SearchService implements SearchService
         return $this->searchClient->searchOrganizers($searchQuery);
     }
 
-    public function addUiTPASLabels(SearchQuery $searchQuery): void
+
+    private function addUiTPASLabels(SearchQuery $searchQuery): void
     {
-        // CUA = internal label for orgs created by Publiq
-        $searchQuery->addParameter(new Query('labels:UiTPAS* OR labels:Paspartoe or labels:CUA'));
+        $labels = array_map(
+            static fn (string $value) => 'labels:' . $value,
+            $this->uiTPASLabelProvider->getLabels()
+        );
+
+        $searchQuery->addParameter(new Query(implode(' OR ', $labels)));
     }
 }
--- a/app/Search/SearchServiceProvider.php
+++ b/app/Search/SearchServiceProvider.php
@@ -6,10 +6,15 @@ namespace App\Search;
 
 use App\Search\Sapi3\Sapi3SearchService;
 use App\Search\Sapi3\SearchService;
+use App\Search\UiTPAS\CachedUiTPASLabelProvider;
+use App\Search\UiTPAS\HttpUiTPASLabelProvider;
+use App\UiTPAS\UiTPASConfig;
 use CultuurNet\SearchV3\SearchClient;
 use CultuurNet\SearchV3\Serializer\Serializer;
 use GuzzleHttp\Client;
+use Illuminate\Contracts\Cache\Repository as CacheRepository;
 use Illuminate\Support\ServiceProvider;
+use Psr\Log\LoggerInterface;
 
 final class SearchServiceProvider extends ServiceProvider
 {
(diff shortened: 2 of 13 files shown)


app/Mails/Smtp/BladeMailTemplateResolver.php
```diff
@@ -20,14 +20,16 @@ public function __construct(
         private ViewFactory $view,
     ) {
         $this->templates = [
-            MailTemplate::INTEGRATION_ACTIVATED->value => 'integration.activated',
+            MailTemplate::INTEGRATION_CREATED->value => 'integration.uitpas.created',
             MailTemplate::ORGANISATION_UITPAS_REQUESTED->value => 'UdbOrganizer.requested',
+            MailTemplate::INTEGRATION_ACTIVATED->value => 'integration.uitpas.activated',
```
LucWollants, line 25: Why is this linked with `integration.uitpas.activated` and not `integration.activated`? It feels like you are reusing the UiTPAS specific mails now also for the normal integrations?

app/Domain/Integrations/Integration.php
```diff
@@ -280,4 +281,15 @@ public function toArray(): array
             'coupon' => $this->coupon ?? null,
         ];
     }
+
+    public function getUdbOrganizerByOrgId(UdbUuid $udbId): ?UdbOrganizer
```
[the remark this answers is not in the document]
LucWollants: You are in the context of an integration here and not in the context of a repository. So the integration id makes no sense, it can only be the organizer id?

config/mail.php
```diff
@@ -38,15 +38,15 @@
...
-            'port' => env('MAIL_PORT', 587),
-            'encryption' => env('MAIL_ENCRYPTION', 'tls'),
-            'username' => env('MAIL_USERNAME'),
-            'password' => env('MAIL_PASSWORD'),
+            'dsn' => env('MAIL_DSN'),
             'timeout' => null,
             'local_domain' => env('MAIL_EHLO_DOMAIN'),
         ],
 
+        'mailpit' => [
```
LucWollants, line 46: I wonder why this is added here? It is the Mailpit API and not DSN

grubolsch: I need 2 different env variables for mailpit:

MAIL_DSN="smtp://127.0.0.1:1025"

This is the "normal" smtp setting, so will be AWS on production.

MAILPIT_DSN="http://127.0.0.1:1025"

This needs be an HTTP protocol link, and is used for API calls in my test with Mailpit. On prod this will be NULL.

https://github.com/cultuurnet/appconfig/pull/1135/files

app/Domain/Integrations/Models/UdbOrganizerModel.php
```diff
@@ -35,7 +36,12 @@ public function toDomain(): UdbOrganizer
     protected static function booted(): void
     {
         self::created(
-            static fn (self $model) => UdbOrganizerCreated::dispatch(Uuid::fromString($model->id))
+            static function (self $model) {
+                UdbOrganizerCreated::dispatch(Uuid::fromString($model->id));
```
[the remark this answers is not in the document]
grubolsch: Indeed, this was the goal.
I did the change for 2 reasons:
1) It is more verbose in meaning
2) I only throw this event when the status of the organizer is pending.

```
    {
        self::created(
            static function (self $model) {
                UdbOrganizerCreated::dispatch(Uuid::fromString($model->id));
                if ($model->status === UdbOrganizerStatus::Pending->value) {
                    UdbOrganizerRequested::dispatch(new UdbUuid($model->id), Uuid::fromString($model->integration_id));
                }
            },
        );
```

app/Domain/Integrations/Integration.php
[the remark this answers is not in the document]
grubolsch: That would not be clear for me?
I would expect `getUdbOrganizerById` to match on the primary key of UdbOrganizer (which is just random in PPF), and not with the UUID that is also know in UDB (the org id).

config/mail.php
LucWollants: So I would use a more clear name like `MAILPIT_API_URL`

app/Domain/Integrations/Integration.php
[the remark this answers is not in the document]
LucWollants: Ok, let's keep the method name!

app/Domain/Integrations/Models/UdbOrganizerModel.php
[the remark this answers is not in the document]
grubolsch: I will add it to my examples in my courses. :-D

tests/UiTPAS/Listeners/SendUiTPASMailsMailpitTest.php
```diff
@@ -108,7 +108,7 @@ public function testMailWasSentIntegrationCreatedWithContacts(
         string $subject
     ): void {
         if (empty($this->mailpitUri)) {
-            $this->markTestSkipped('MAILPIT_DSN is not set, skipping email tests.');
+            $this->markTestSkipped('MAILPIT_API_URL is not set, skipping email tests.');
```
LucWollants, line 111: I guess another hard fail is needed here. Otherwise, you could get false positives when `MAILPIT_API_URL` is missing and not executed?

app/Mails/Smtp/BladeMailTemplateResolver.php
grubolsch: I checked all templates, the activated template is indeed shared. It is currently not sent yet, but will be used once we remove the Mailjet integration, which I am working on this week.

PR #1938, opened 2025-08-04
PPF-700 Moved uitpas event handlers

grubolsch:
### Fixed

---

### DIFF
--- a/tests/UiTPAS/Listeners/NotifyUdbOrganizerRequestedTest.php
+++ b/tests/UiTPAS/Listeners/NotifyUdbOrganizerRequestedTest.php
@@ -5,16 +5,16 @@ declare(strict_types=1);
 namespace Tests\UiTPAS\Listeners;
 
 use App\Domain\Integrations\Environment;
-use App\Domain\Integrations\Events\UdbOrganizerCreated;
+use App\Domain\Integrations\Events\IntegrationActivationRequested;
 use App\Domain\Integrations\IntegrationType;
 use App\Domain\Integrations\Repositories\IntegrationRepository;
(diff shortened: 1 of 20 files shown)


tests/UiTPAS/Listeners/NotifyUdbOrganizerRequestedTest.php
```diff
@@ -176,6 +161,43 @@ public function test_it_does_not_message_approved_organizers(): void
         $this->messageBuilder->expects($this->never())
             ->method('toMessageWithOrganizer');
 
-        $this->listener->handle($event);
+        $this->listener->handleIntegrationActivationRequested($event);
+    }
+
+    public function test_it_sends_mail_when_udb_organizer_is_requested(): void
```
LucWollants, line 167: Is it not the notifier we want to test here, and not the sending of mails?

LucWollants (approved):
Looks good! I did not test if all the scenarios still work though.

tests/UiTPAS/Listeners/NotifyUdbOrganizerRequestedTest.php
grubolsch: I am sorry, I do not understand your remark. can you clarify?

LucWollants: I mean in the context of this test file, we are mainly interested in sending the Slack notification. You are testing an extra side effect of dispatching the event, which is also sending mails, but that is not the goal of `NotifyUdbOrganizerRequested`

PR #1943, opened 2025-08-07
PPF-700 Disable mailjet mails for uitpas

grubolsch:
### Changed
- The new code for UiTPAS will sent mails by SMTP. Because of this currently 2 mails are sent as duplicates, with slightly different copy. This code disables the Mailjet mails IF the feature flag for UiTPas is toggled.

---


app/Domain/Mail/MailManager.php
```diff
@@ -43,13 +43,23 @@ public function sendIntegrationCreatedMail(IntegrationCreatedWithContacts $event
     {
         $integration = $this->integrationRepository->getById($event->id);
 
+        if ($integration->type === IntegrationType::UiTPAS && config(UiTPASConfig::AUTOMATIC_PERMISSIONS_ENABLED->value)) {
```
LucWollants, line 46: I think the check for the feature flag has no added value and is not even needed.

LucWollants (approved):

app/Domain/Mail/MailManager.php
grubolsch: removed the flag

PR #1952, opened 2025-08-11
PPF-677-Fix alert visibility when credentials expiration date is null

vhande:
### Fixed
- Alert visibility when credentials expiration date is null

---
Ticket: https://jira.uitdatabank.be/browse/PPF-677

### DIFF
--- a/resources/ts/Components/Integrations/Detail/CredentialsLegacyAuthConsumers.tsx
+++ b/resources/ts/Components/Integrations/Detail/CredentialsLegacyAuthConsumers.tsx
@@ -35,12 +35,14 @@ export const CredentialsLegacyAuthConsumers = ({
   oldCredentialsExpirationDate: oldCredentialsExpirationDateString,
 }: Props) => {
   const { t, i18n } = useTranslation();
-  const oldCredentialsExpirationDate = new Date(
-    oldCredentialsExpirationDateString
-  );
-  const timeLeft = formatDistanceToNow(oldCredentialsExpirationDate, {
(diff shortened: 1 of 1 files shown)


resources/ts/Components/Integrations/Detail/CredentialsLegacyAuthConsumers.tsx
```diff
@@ -35,12 +35,21 @@ export const CredentialsLegacyAuthConsumers = ({
   oldCredentialsExpirationDate: oldCredentialsExpirationDateString,
 }: Props) => {
   const { t, i18n } = useTranslation();
-  const oldCredentialsExpirationDate = new Date(
-    oldCredentialsExpirationDateString
-  );
-  const timeLeft = formatDistanceToNow(oldCredentialsExpirationDate, {
-    locale: languageToLocale[i18n.language],
-  });
+  const oldCredentialsExpirationDate = oldCredentialsExpirationDateString
```
LucWollants, line 38: I'm not sure about the naming of this variable, because the name was a reason for the confusion. Keys basically have no expiration date. It is a grace period for the switch from v1 to v2 keys, which is calculated from the moment the integrator requested to upgrade from v1 to v2 keys. Maybe a name like `keyVisibleUntil` says more?

bertramakers: I think this is a good suggestion because it clarifies that this property can influence the `keyVisibility`.
Other than that I like the overall structure of the code with the clear variable names like `isExpired` 👍

